PR #1, opened 2019-11-13
Initial performer prototype

WithoutPants:
Basic prototype of a performer database. Based on a rewrite of the stash codebase. Conforms mostly to the proposed graphql schema with minimal changes (which I'll have to update the original with). 

Allows creation, modification, deletion and basic querying of performers. Not all filter options are implemented.

It's currently named `stashdb` and runs on port 9998. On initial run, it creates a config file with api keys for reading and writing. This api key must be set as a value for `ApiKey` request header. The only interface is the graphql API currently (via `/graphql`), but the playground is enabled at `<server>:9998/playground`.

To build: 

- Run `go mod vendor` to copy external libraries to the vendor directory, since I'm not currently storing them in the repository. 
- Run `make generate` to generate the graphql files and the packr2 files for database migrations.
- Run `make build` to build the binary.

Config file and database are created in the working directory when stashdb is started.


pkg/database/database.go
```diff
@@ -0,0 +1,84 @@
...
+		Box:        migrationsBox,
+		Migrations: source.NewMigrations(),
+	}
+
+	databasePath = utils.FixWindowsPath(databasePath)
+	s, _ := WithInstance(packrSource)
+	m, err := migrate.NewWithSourceInstance(
+		"packr2",
+		s,
+		fmt.Sprintf("sqlite3://%s", "file:"+databasePath),
```
bnkai: should be `fmt.Sprintf("sqlite3://%s", databasePath),` probably copy paste left from sqlx invocation?

Leopere:
I wonder if for storage efficiency we should have all database entries be an integer which would refer to a database of tags.

pkg/database/database.go
WithoutPants: Why are we dropping the `file:` prefix?

bnkai: Edit . The specific panic happens when i compile with go 1.13 , `go version go1.13.4 linux/amd64` to be exact 
with 1.11.5 everything seems fine

WithoutPants: It appears that the presence of `file:` meant that the subsequent path must be an absolute path. Since stashdb was by default passing a relative path to the database file, it was causing a panic on Linux. Removing the `file:` directive fixed the panic and has no apparent issues on Windows.

PR #5, opened 2019-11-27
Frontend for stash-box

ghost:
CRUD UI for the stash_prototype go backend.


Leopere:
Just for reference its currently located at http://whater.stashapp.cc:8080/ and https://whater.stashapp.cc, I might consider throwing up a Traefik reverse proxy in front of it with TLS provisioning so that we have compatibility-tested that way as well.

PR #7, opened 2019-11-30
 Change type of IDs from int64 to uuid 

ghost:
Stored as native uuids in postgres, and binary blobs in sqlite. Should be more or less indistinguishable in terms of performance and functionality.


pkg/models/model_performer.go
```diff
@@ -298,15 +308,6 @@ func (p *Performer) CopyFromCreateInput(input PerformerCreateInput) error {
...
 
-	if input.Image != nil {
-		var err error
-		p.Image, err = p.TranslateImageData(input.Image)
-
-		if err != nil {
-			return err
-		}
-	}
-
```
WithoutPants: This looks unrelated to the uuid change. Why was this removed from here?

pkg/models/querybuilder_sql.go
```diff
@@ -37,22 +38,18 @@ func handleStringCriterion(column string, value *StringCriterionInput, query *da
...
 
-	if err != nil {
-		return 0, err
-	}
-
-	return result.LastInsertId()
+	return err
 }
 
 func insertJoins(tx *sqlx.Tx, table string, joins interface{}) error {
```
WithoutPants: Since `insertObject` no longer returns the id, I think we can rename this to `insertObjects`.

WithoutPants (changes requested):
One minor change and a question.

pkg/models/model_performer.go
ghost: My bad, fixed.

pkg/models/querybuilder_sql.go
ghost: It still only inserts a single object so I'm not quite sure why we would change it?

WithoutPants: This refers to line 55 - to rename `insertJoins` to `insertObjects`.

ghost: Oh, gotcha, the diff confused me. I've updated the name.

WithoutPants (approved):

PR #10, opened 2019-12-18
Align backend and frontend

ghost:
* Replace url field on scene table with scene_urls table.
* Remove database stored images. Slow, and not particularly useful for our purposes.
* Remove wasFieldIncluded. Didn't work with sub-properties, so the test was always false. Unfortunately there's no particularly good solution in gqlgen besides enumerating properties on the context, so I would suggest just leaving it for now. On the frontend I'm currently always doing full updates, so it's not really an issue.
* ~~Replace `?` with `$1` in raw queries since postgres doesn't support the question mark operator for parameter replacements. SQLite supports both kinds.~~
* Fix sqlx query placeholder rebinding
* Raise the max row limit on queries to 1000. We need to be able to query all studios, and in my test set there's currently somewhere around 500.
* Add `NULLS LAST` to ORDER clause. Postgres sorts NULLs first by default when ordering DESC, which screws with optional date ordering. SQLite added support for NULLS LAST/FIRST in the latest version (3.30.1), though I'm not sure if go-sqlite3 had added support for it yet.


WithoutPants:
> * Remove database stored images. Slow, and not particularly useful for our purposes.

This has been removed and not replaced. Storing images is supposed to be included in the scope of stash-box. What's the implementation idea of this going forward?

> * Remove wasFieldIncluded. Didn't work with sub-properties, so the test was always false. Unfortunately there's no particularly good solution in gqlgen besides enumerating properties on the context, so I would suggest just leaving it for now. On the frontend I'm currently always doing full updates, so it's not really an issue.

This may become necessary once we start doing edits, but it's fine to leave off for now.

> * Replace `?` with `$1` in raw queries since postgres doesn't support the question mark operator for parameter replacements. SQLite supports both kinds.

This was already covered in `sqlDialect.SetPlaceholders`. There are still other places that use `?`, so I'm not sure why we need to convert specific queries. I left it as `?` because I didn't want to have to keep track of placeholder indexes. Prefer maintaining `?` and using `SetPlaceholders` to convert, since by that time we have the entire sql query string to convert.

ghost:
> > * Remove database stored images. Slow, and not particularly useful for our purposes.
> 
> This has been removed and not replaced. Storing images is supposed to be included in the scope of stash-box. What's the implementation idea of this going forward?
> 
The replacement would be the url fields. Personally I don't think we should be dealing with hosting images right now for legal reasons, but even if we were to do that, images should be stored on a cdn, and the database should only concern itself with urls to the resource. Pulling images from the database through graphql is going to kill performance for a whole host of reasons, for instance blocking metadata loading, a complete lack of client caching and no parallell image loading.

> 
> > * Replace `?` with `$1` in raw queries since postgres doesn't support the question mark operator for parameter replacements. SQLite supports both kinds.
> 
> This was already covered in `sqlDialect.SetPlaceholders`. There are still other places that use `?`, so I'm not sure why we need to convert specific queries. I left it as `?` because I didn't want to have to keep track of placeholder indexes. Prefer maintaining `?` and using `SetPlaceholders` to convert, since by that time we have the entire sql query string to convert.

Ah, I didn't see the setPlaceholder calls until now. In that case it's probably a bug somewhere, I can take a look tomorrow and see if I can figure out why the placeholders aren't being replaced.

ghost:
I've reverted the query placeholder change. The issue was that sqlx doesn't automatically rebind placeholders, so we have to manually call `Rebind` any time a query is executed directly in a transaction.

WithoutPants (approved):

WithoutPants:
Looks like `Rebind` effectively replaces `SetPlaceholders`. That will have to be a future refactor. For now, if this works, then it's good enough for me.

PR #12, opened 2020-01-12
Fix integration tests

WithoutPants:
- Fix test compilation errors
- Remove the integration tests that perform partial modify operations, since these are no longer supported.
- Move `NULLS LAST` into dialect function. This isn't supported in the current golang sqlite library. There is probably a workaround available, but I'll leave it for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ TODO
 * `make test` - Runs the unit tests
 * `make it` - Runs the unit and integration tests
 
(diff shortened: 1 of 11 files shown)


WithoutPants:
I've added the ability to run the integration tests against a postgres schema via the `POSTGRES_DB` environment variable. Note that this drops all of the tables before and after the test, so make sure you set it to a separate test schema and not anything with data you need.

PR #13, opened 2020-01-20
Add User management

WithoutPants:
Adds user management, via graphql query only currently.

Changed so that, if the root user is not created, then a root user is generated with a random generated password and an API key. The password and api key is output to stdout, and not logged. The root user name and roles cannot be changed. The root user also may not be deleted.

Changed so that the user API key is required to interact with the system, and the old api key settings are removed. API key is a generated JWT, including the user ID only. The signing key is generated if not present and added to the configuration.

Role authorisation is implemented. The following rules are in place:

- Any query and mutation of users requires the `ADMIN` role. This will be changed in future to relax this somewhat, to allow the current user to query/mutate their user object, and to query some user fields (explicitly excluding api keys).
- Mutating of any other object requires the `MODIFY` role.
- Reading of any other object requires the `READ` role.

The other roles (`VOTE` and `EDIT`) are currently unused. All non-`READ` roles imply `READ`, meaning that the user does not need an explicit `READ` role if they have one of the others. Similarly, `ADMIN` implies all roles.

Password policies are currently hardcoded, and include the following rules:

- length must be between 8 and 64 characters, inclusive
- must have at least 5 unique characters
- must not match the username or email of the user, or appear in the top 10,000 common passwords list

Added `/login` and `/logout` end points. Login accepts form-based `username` and `password` fields.

I've briefly tested against the frontend and updated the `README` accordingly.

This change also includes some other miscellaneous fixes:
- change the `packr` call to `Find` to remove the deprecation message
- add `gitattributes` to correctly set the line ending on the UI files, to remove an eslint warning
- change `it` make target to not cache the results

Future changes:
- ~~add sessions, login and logout~~
- ~~add `regenerateApiKey` graphql mutation~~
- ~~add `changePassword` graphql mutation (for current user only)~~
- add `resetPassword` graphql mutation (for current user only)
- ~~fix authorisation for user queries and mutations~~


pkg/manager/user.go
```diff
@@ -0,0 +1,393 @@
...
+
+func UserCreate(tx *sqlx.Tx, input models.UserCreateInput) (*models.User, error) {
+	var err error
+
+	UUID, err := uuid.NewV4()
+	if err != nil {
+		return nil, err
+	}
+
+	// Populate a new studio from the input
```
ghost: Studio - > user, couple more further down.

ghost (approved):
This is really great stuff!

We should maybe consider sticking the sessions in a database table so we can invalidate on demand, but no need for that now.

Other than the copy pasted comments I have no complaints.

pkg/manager/user.go
WithoutPants: Fixed

Leopere:
I know I'm late to the party here but 8 characters is far beneath the minimum when it comes to secure passwords technically speaking 16/18 chars is unbruteforceable and below that its just a matter of attempts.

PR #19, opened 2020-06-12
Omnibus stashdb changes

ghost:
Splitting up the various PRs proved a bit too difficult due to later commits having messy dependencies, so I've just cleaned it up as one big PR instead.

The changes include:
* Full text search table. Same as #11 with tweaks to yield better results.
* Frontend support for auth and user management. Same as #15 generally speaking.
* Tags/fingerprints support. Same as #16.
* Add duration to fingerprints. Submitted by the tagger and used when scanning. Also very useful for identifying bad fingerprints (40m fingerprint on 20m scene, etc.).
* Add separate image tables for all entities. I initially tried putting images in the url tables, but it lead to a lot of boilerplate to separate in the frontend. I also needed the width/height for sorting and resize redirects, and it was messy adding those columns to the url tables. In stashdb I have some additional logic in the handlers related to cdn resizing, but I've omitted it here.
* Update styles to match stash.
* Ran everything through prettier. Turned out to be a bad idea since it made it impossible to separate any changethat came after the reformat, but not much to be done about it now.

In general the frontend stuff is a bit of a mess and I want to clean things up and simplify, as well as switch the build system to match stash, but I'll leave that for later.


graphql/schema/types/misc.graphql
```diff
@@ -25,6 +25,9 @@ enum SortDirectionEnum {
 type URL {
   url: String!
   type: String!
+  image_id: String
```
WithoutPants: Should this be `ID`?

go.mod
```diff
@@ -4,21 +4,48 @@ go 1.12
```
WithoutPants, line 4: `sql.NullInt32` was introduced in go version 1.13. The version in go.mod needs to be updated.

frontend/package.json
```diff
@@ -81,13 +86,15 @@
...
     "bootstrap": "^4.3.1",
-    "graphql": "^14.5.8",
-    "graphql-tag": "^2.10.1",
+    "graphql": "^14.6.0",
+    "graphql-tag": "^2.10.3",
     "i18n-iso-countries": "^4.3.1",
     "lodash": "^4.17.15",
     "query-string": "^6.9.0",
-    "react-bootstrap": "^1.0.0-beta.14",
+    "react-bootstrap": "^1.0.2",
```
WithoutPants: This doesn't match any known versions. Running `yarn install` gives the following output:
```
warning Lockfile has incorrect entry for "react-bootstrap@^1.0.2". Ignoring it.
Couldn't find any versions for "react-bootstrap" that matches "^1.0.2"
? Please choose a version of "react-bootstrap" from this list: 
```
I chose 1.0.1 because that was what was in the lockfile. I was unable to run `yarn install --frozen-lockfile`. I think this needs to be changed to `1.0.1`.

frontend/package.json
```diff
@@ -51,16 +54,18 @@
     "import-glob-loader": "^1.1.0",
     "jest": "^24.9.0",
     "node-sass": "^4.13.0",
+    "prettier": "2.0.2",
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-hot-loader": "^4.12.16",
```
WithoutPants: Don't these belong in `devDependencies`?

WithoutPants (changes requested):
I haven't had much luck building this.

I first had to delete the UI `yarn.lock` and `node_modules` then run `yarn install`, otherwise I'd get errors due to `@types/react` resolving twice ([see here](https://stackoverflow.com/questions/52399839/duplicate-identifier-librarymanagedattributes)).

I'm now getting the following error when running `yarn build`:

```
ERROR in ./src/components/sceneForm/SceneForm.tsx 113:36
Module parse failed: Unexpected token (113:36)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
|       details: data.details,
|       studio_id: data.studioId,
>       performers: (data.performers ?? []).map(performance => ({
|         performer_id: performance.performerId,
|         as: performance.alias
 @ ./src/components/sceneForm/index.ts 10:0-36 11:17-26
 @ ./src/components/sceneEdit/SceneEdit.tsx
 @ ./src/components/sceneEdit/index.ts
 @ ./src/App.tsx
 @ ./src/index.tsx

ERROR in ./src/components/fragments/Icon.tsx 20:16
Module parse failed: Unexpected token (20:16)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
|   icon: icon,
|   className: `fa-icon ${className}`,
>   color: color ?? "white"
| });
|
 @ ./src/components/fragments/index.ts 8:0-41 8:0-41
 @ ./src/components/home/Home.tsx
 @ ./src/components/home/index.ts
 @ ./src/App.tsx
 @ ./src/index.tsx

ERROR in ./src/components/fragments/LoadingIndicator.tsx 26:12
Module parse failed: Unexpected token (26:12)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
 * ./node_modules/ts-loader/index.js
You may need an additional loader to handle the result of these loaders.
| }, "Loading...")), React.createElement("h4", {
|   className: CLASSNAME_MESSAGE
> }, message ?? "Loading..."));
|
| const _default = LoadingIndicator;
 @ ./src/components/fragments/index.ts 6:0-65 6:0-65
 @ ./src/components/home/Home.tsx
 @ ./src/components/home/index.ts
 @ ./src/App.tsx
 @ ./src/index.tsx
<snip>
```

frontend/package.json
ghost: Package has been cleaned up now, so should mostly match stash.

frontend/package.json
ghost: Not sure how that happened, fixed.

go.mod
ghost: Not sure it makes any difference since the database value is the same anyway. I just switched it to NullInt64 instead.

graphql/schema/types/misc.graphql
ghost: Shouldn't be there at all actually since image is now it's own entity. Fixed.

ghost:
Did a poor job testing this it seems. Anyway, I switched the frontend project to create-react-app so it largely matches stash. I also upgraded typescript, which exposed a whole raft of null errors. I haven't tested the changes extensively, but it should install and run fine.

Note that to log in you either have to run the release build from the go binary and log in as normal, or embed the apikey in `frontend/.env`  and run `yarn start`. There's no way to log in in dev mode since cookies need to be on the same domain/port, iirc.

Also, to run the search migration you first need to run `create extension pg_trgm;` on the stash-box database as a superuser. I considered putting it in the migration, but that would necessitate running the whole thing as superuser which sounds not good.

WithoutPants:
Compile errors:
```
pkg\models\model_studio.go:71:6: url.ImageID undefined (type URL has no field or method ImageID)
pkg\models\model_studio.go:72:6: url.Height undefined (type URL has no field or method Height)
pkg\models\model_studio.go:73:6: url.Width undefined (type URL has no field or method Width)
```

ghost:
@WithoutPants Ugh, forgot to remove those. Fixed.

WithoutPants:
I needed to upgrade logsrus to 1.6 to eliminate a panic on Windows.

WithoutPants:
Get this in the integration tests:
`pkg\api\scene_integration_test.go:214:60: cannot use fingerprint (type models.FingerprintInput) as type models.FingerprintQueryInput in argument to s.testRunner.resolver.Query().FindSceneByFingerprint`

Can you please update the `README.md` file with the information from your last comment?

Please update logsrus to 1.6.0.

As far as my testing is concerned, I got to the point where I could get the login page showing up (using `yarn start`). Logging in was apparently succcessful - the login endpoint returned `200` and a session cookie was returned, but the subsequent requests seem to lose the cookie and I was returned to the login page again. 

I'm prepare to merge this in if the integration tests check out, but I think the UI needs more kinks worked out of it.

ghost:
The logsrus crash is weird, stash seems to be mandating 1.4.2. Anyway, I updated it to 1.6. I also fixed a bunch of bugs and rewrote the frontend graphql connection to match stash, so it's no longer necessary to configure it manually. As long as the apikey is configured it should run correctly in dev mode.

I've deployed this build on stashdb.org and I haven't seen any issues so far at least.

WithoutPants:
I've submitted a PR for your branch, rather than pushing directly.

ghost:
Thanks, I managed to resolve the cookie issue after a lot of headbanging.  I've verified that it works in dev, and in production. Definitely makes it a lot smoother to develop.

frontend/apollo.config.js.shadow
```diff
@@ -2,7 +2,7 @@ module.exports = {
     client: {
         service: {
             name: 'stashdb',
-            url: 'http://localhost:9998/graphql',
+            url: 'https://stashdb.org/graphql',
```
WithoutPants, line 5: What does this do?

frontend/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Stash",
```
WithoutPants, line 2: This should be changed at some point.

WithoutPants (approved):

PR #20, opened 2020-06-28
Fix bugs and allow for unsetting performer metadata.

ghost:
* Allow unsetting performer properties. This involved setting `sqlGenKeys` partial to true in `updateObjectByID` so it sets zeroed values, submitting zeroed birthdate/measurements when applicable, and returning `nil` when sqliteDate string is empty.
* Show performer disambiguation string in performer names.
* Fix eye/hair color and breast type being always required when editing.
* Resort country list and add `unknown` option for unsetting purposes.
* Set form cancel buttons to go back rather than a fixed url.
* Add create buttons to scene/performer/studio/tags, and tweak styling.
* Fix search field and prevent it from searching for empty strings and showing the result dropdown before a search is finished.
* Show full country name in performer name rather than ISO.

Most of this stuff is relatively straighforward. The exception is the change to updateObjectByID. I *think* it's fine since the frontend always submits the entire object anyway, and I don't see any other way of allowing unsetting properties.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -60,7 +60,7 @@
     "bootstrap": "^4.4.1",
     "graphql": "^14.6.0",
     "graphql-tag": "^2.10.3",
-    "i18n-iso-countries": "^4.3.1",
+    "i18n-iso-countries": "^6.0.0",
     "lodash": "^4.17.15",
     "query-string": "^6.9.0",
     "react": "^16.12.0",
--- a/frontend/src/components/fragments/PerformerName.tsx
+++ b/frontend/src/components/fragments/PerformerName.tsx
@@ -1,0 +1,36 @@
+import React from "react";
+import { Performer_findPerformer as Performer } from "src/definitions/Performer";
+
+interface PerformerNameProps {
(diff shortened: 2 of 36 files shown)


WithoutPants:
Integration test failures:

```
go test -tags=integration -count=1 ./...
--- FAIL: TestUpdateScene (0.01s)
    scene_integration_test.go:330: Error updating scene: Error updating Scene: pq: column "url" of relation "scenes" does not exist
--- FAIL: TestUpdateTag (0.00s)
    tag_integration_test.go:139: Name mismatch:  != 0xc00055c160
--- FAIL: TestUpdateUserName (0.07s)
    user_integration_test.go:52: Email mismatch:  != user-5@example.com
--- FAIL: TestUpdateUserPassword (0.16s)
    user_integration_test.go:227: Error updating user: Error updating User: pq: duplicate key value violates unique constraint "users_email_key"
```

ghost:
@WithoutPants Should be resolved now. A leftover URL field on the scene model, and an oversight related to the change to `updateObjectByID` which was causing updates to wipe fields that weren't set.

PR #21, opened 2020-08-08
Add editing support for tags

ghost:
This PR adds full editing support for tags based on the backend work from @WithoutPants. The frontend is somewhat beta but should be serviceable. Allows for creation of insert/modify/destroy/merge edits, and immediate application/cancellation for admins.


graphql/schema/types/edit.graphql
```diff
@@ -45,19 +51,30 @@ type Edit {
     id: ID!
     user: User!
     """Object being edited - null if creating a new object"""
-    target: EditTarget!
+    target: EditTarget
     target_type: TargetTypeEnum!
     """Objects to merge with the target. Only applicable to merges"""
-    merge_sources: [EditTarget]
+    merge_sources: [EditTarget]!
```
WithoutPants: I think this should be `[EditTarget!]!`.

WithoutPants (changes requested):
Don't know if this a beta UI issue, but when I merged some tags into another, it didn't take the scenes with it. I could navigate to the deleted tag and see its scenes, but those scenes didn't appear in the merged tag.

Can't see a way to look at pending creation edits too.

As far as I can tell, the server side looks alright.

graphql/schema/types/edit.graphql
ghost: True, fixed.

ghost:
I realized last night that I'd actually forgotten to implement the sceneTag move. I think maybe I was considering letting the redirect do the job, which would work but would likely be quite slow with time. Either way, I've fixed it so all sceneTags will be updated to the new ID when merged.

Tag creation edits are only visible from the main edits component since the tag doesn't exist yet. I've added a link in the main menu to it since it seems I forgot.

ghost:
Made a small change to show merge edits in merged tags, so it's possible to follow the editing history.

pkg/database/migrations/postgres/5_edits.up.sql
```diff
@@ -0,0 +1,63 @@
...
+  "user_id" uuid not null,
+  "operation" varchar(10) not null,
+  "target_type" varchar(10) not null,
+  "data" jsonb,
+  "votes" integer not null default 0,
+  "status" varchar(20) not null,
+  "applied" boolean default FALSE not null,
+  "created_at" timestamp not null,
+  "updated_at" timestamp not null,
+  foreign key("user_id") references "users"("id")
```
WithoutPants: I think this might need some further thought. Need to consider that users may be removed.

WithoutPants:
After editing a tag description and accepting it, the edit screen is incorrect. In this example, I add the description, so the previous value should be empty:

image

In the following example, I changed the description from empty to `edit-1`, then after applying, I changed the description to `edit-2`. 

image

The "Modify Tag" edit entry in the edits list needs to include a link to the tag being edited.

I think that this will need a suite of integration tests in the `api` package. Let me know if you want me to contribute those.

pkg/database/migrations/postgres/5_edits.up.sql
ghost: Good point. I think it should suffice to make the column nullable, and set `ON DELETE SET NULL`, so users can be freely deleted without taking edit history down with them.

ghost:
> After editing a tag description and accepting it, the edit screen is incorrect. In this example, I add the description, so the previous value should be empty:

Yeah, it's a display issue. The old data is saved in the edit, but the resolver always returns the current data. I'll get that fixed.

> 
> image
> 
> In the following example, I changed the description from empty to `edit-1`, then after applying, I changed the description to `edit-2`.
> 
> image
> 
> The "Modify Tag" edit entry in the edits list needs to include a link to the tag being edited.

True, will do.

> I think that this will need a suite of integration tests in the `api` package. Let me know if you want me to contribute those.

Yeah, I discovered a bunch of bugs yesterday that would've been trivially uncovered by tests. I'll get started on testing the applyEdit in particular and let you know how it goes.

ghost:
Should have more or less coverage of the new edit endpoints now. The exception is the redirect table which isn't actually used for anything yet. Will be more relevant once we move on to other entities.

ghost:
I've deployed the PR on stashdb.org. No issues that I've noticed, so as far as I'm concerned it's good to merge.

WithoutPants (approved):

PR #25, opened 2020-08-21
Add dataloaders

ghost:
This PR adds dataloaders for all commonly used objects. The motivation for adding dataloaders is the N+1 problem inherent to GraphQL: https://medium.com/the-marcy-lab-school/what-is-the-n-1-problem-in-graphql-dd4921cb3c1a

The dataloaders are generated by dataloaden (https://gqlgen.com/reference/dataloaders/). The advantage is that in addition to batching queries, it only fetches each entity once and caches the result, so when loading a performer page the performer is only loaded once, rather than once for each scene.

The performance improvements have been pretty hefty, from 50-200% depending on payload size. I've already deployed it on stashdb.org since the improvements were noticeable.

There's quite a bit of boilerplate due to the need to aggregate and sort the query results, which gets tricky to do in a generic manner. I'll try to think if it's possible to generalise it, but might not be worth it.


WithoutPants:
Can you please include makefile targets to regenerate the dataloaders?

ghost:
Sure, bit noisy but should work.

WithoutPants (approved):

Issue #17, opened 2020-04-22
[RFC] Creating a master list for tags


AdultSun:
Hello everyone! This is a long post, so hopefully I'm not overstepping my bounds here. I sent a rough pitch on how I thought tags could work to Stash (meaning the original developer) about a year ago. Basically, I shared the list of tags I use within my own personal, local database thinking it would be useful during development. Since work here seems to have become much less theoretical than those early days, I thought it could be worth sharing again.

(Also, my coding experience is extremely limited to the point of non-existence, so bear with me if I don't understand a lot of the more technical language or if I over-explain something.)

The premise I'm starting from is that for the tags contained in StashBox to be as useful and functional as possible, we should have a relatively static master list of possible tags. These tags should have clear definitions so both users and contributors know what everything means. Without a level of internal consistency to these tags, the accuracy and reliability of our data will suffer.

Stash saved and shared the outline/list/dictionary I sent him before. [You can find the whole thing here](https://gist.github.com/StashAppDev/cf71bd17cfd4397c16415155bf72fe07), but because it's so massive to the point of being embarrassing I'll give a brief synopsis as well so nobody feels like they have to scroll through all of it.

Because I have such a large list of tags, the only way to keep it manageable was to group everything together. I'm not sure how everybody feels about how tags should be structured and implemented, so for now it can just be seen as a convenient way of organizing my list so it can be more easily read and understood. I also can provide screenshots of my database in case anyone is curious what this system looks like in practice. I assure you it's much neater than the massive document above would imply.

## Synopsis
I have everything organized into three broad sections first, with smaller categories underneath each of them. 

**People Tags** contains tags that describe the appearance of a performer and may be listed on either a performer's individual profile page or to describe the appearance of performers within a particular scene. 
- Categories include *Age Group*, *Hair Color*, and *Height*. 
- Tags include *Teen*, *Blonde*, and *Petite*.

**Scene Tags** contains tags that describe the make-up of a particular scene/movie. 
- Categories include *Group Makeup*, *Relationship*, and *Location*. 
- Tags include *Couple*, *Girlfriend*, and *Classroom*.

[...]

PR #28, opened 2020-09-04
Add Tag Categories

ghost:
Categories and groups are based on the proposal by @AdultSun in #17. Categories are intended to be relatively static, so admin editable only.

Most of the code is boilerplate, same as previous entities. The categories can be accessed in `/categories` which is linked from the tag list.


WithoutPants:
Doesn't look like its possible to unset category in tag.

ghost:
> Doesn't look like its possible to unset category in tag.

True, I've added it in now.

PR #31, opened 2020-11-10
User registration

### LINKED ISSUE #29: [RFC] User registration and invite system
# Introduction

This RFC covers the user registration and activation process and the user invite system.

# Configuration

Propose to add the following configuration keys:

* `require_invite` [`true`] - if false, allows new users to create accounts without needing an invite key
* `activation_expiry` [2 hours] - time in seconds from when an activation event is generated (new user account, reset password requests) before the activation key expires
* `email_cooldown` [5 minutes] - time in seconds before a second activation email may be generated - to prevent spamming

# New user process

1. New user enters email address and (if `require_invite` is true) a valid invite key.
2. System sends an email with a link to activate the account. The link includes a generated activation key.
3. User clicks the link and the system shows the activation page. User enters their email address (to verify), the password for the new account and hits submit.
4. System verifies the activation key and creates the account, setting the password.

If the user clicks the link after the `activation_expiry` period has passed, then the system reports that the activation link is no longer valid.

An invite key may only be used once. When the activation link is sent out, the invite key is considered used. The key may only be used again if the activation expiry has passed and the account was not created. The same key may be used to resend the activation email to the _same_ email address, but only after the `email_cooldown` time period has passed.

Invite keys will be associated with the user that generated them. Accounts created by invite keys will be associated with the account that created the invite key. This association will only be viewable by users with the `ADMIN` role.

# Invite system

Users may be granted any number of invite tokens, these may be used to generate invite keys. As mentioned above, an invite key may only be used once. A user that generated an unused invite key may rescind it and reuse the token to generate another key.

Two new roles will be created: `MANAGE_INVITES`, and `INVITE`.

A user with the `INVITE` role can generate new invite keys and rescind unused invite keys that they generated. They do not require invite tokens to generate invite keys.

A user with the `MANAGE_INVITES` roles may grant and rescind invite tokens for users to generate invite keys, and may rescind any invite keys. Invite keys rescinded in this manner do not refund the invite token.

Token management should be logged, including the user that performed the operation.

# Interfaces

`newUser` - requires email address and optionally an invite key. Creates a pending user and sends email containing activation key.
`activateNewUser` - requires email address, activation code and password to set. Validates the activation code against the email address and creates the new user
`generateInviteCode` - use an invite token (if applicable) to generate an invite key. Returns the invite key.
`rescindInviteCode` - invalidates an invite key, refunding the invite token if the key was generated by that user. 
`grantInvite` - increments the invite tokens for a user.
`repealInvite` - decrements the invite tokens for a user.

I will also include `resetPassword` - which will send an email with an activation key to reset an existing users password, and change `changePassword` to optionally accept the activation key to set the password.

WithoutPants:
Resolves #29 

Adds all of the changes described in #29. Adds a very basic set of UI pages that allow granting and revoking invite tokens, generating and rescinding invite key, user registration and activation, and password recovery.


ghost:
Works great! Feel free to merge it, I'll update the migration in the tag category PR.

PR #33, opened 2020-11-19
Image local storage and editing

WithoutPants:
Changes to support uploading of image data when creating Images. 

Adds a new configuration key (`image_location`), which indicates where to store the images. Images are stored as their checksum in the specified directory.

Adds the ability to add and remove images to Scenes, Performers and Studios.


pkg/api/resolver_model_image.go
```diff
@@ -12,7 +13,14 @@ func (r *imageResolver) ID(ctx context.Context, obj *models.Image) (string, erro
 	return obj.ID.String(), nil
 }
 func (r *imageResolver) URL(ctx context.Context, obj *models.Image) (string, error) {
-	return obj.URL, nil
+	// if checksum is populated, then generate a URL, otherwise use the URL
+	if obj.Checksum.Valid {
```
ghost, line 17: I think checksums might end up being useful for cdn storage as well, so I think we should have a separate config option to explicitly check what file storage to use and return the appropriate url, rather than looking at the presence of checksums. It can fallback to file if the config option isn't set.

ghost (approved):
Looks great! I can add the config option for which image backend to use when I add the cdn storage.

pkg/api/resolver_model_image.go
ghost: I've pushed a commit that makes the checksum/width/height not nullable. It's very useful for deduplicating, so no reason to not always include it. I've left the models alone, will update in a separate pr for s3 storage.

PR #35, opened 2020-12-10
Add performer editing

ghost:
Adds editing of performers, similar to the existing tag editing. It allows for creating/deleting/merging entities, as well as editing and all fields, with the exception of image addition, and performer url addition/removal. Performer urls for instagram, twitter, etc. will be handled in a separate PR since it requires additional work to support custom url types.

I've remade the performer form to be less bad, and added support for edit comments to both tags and performers. There is no way to add additional comments to an edit currently, but I don't think that's particularly urgent.


WithoutPants:
Can you resolve the conflicts in this PR? I'm not confident doing it myself while reviewing.

ghost:
PR has been rebased on dev. Since I had to retrofit the image uploading into the performer edit form I spruced it up a bit as well.

Everything seems to work except some oddities around deletions, but it's hard to say what's a real error and what's my screwed up dev database.

 The only issue I'm aware of is the additional check I added to `IsUnused` that checks if an image is pending addition through an edit. This query parses the edit json to get the image ids, which is likely quite slow, and not indexable. We may have to add a join to the image for edits that add images, but it runs fine for now.

WithoutPants:
Deleted performers still appear in the search results when entering the performer name in the search box.

I noticed some updating issues when I rejected an edit (the status didn't seem to update). I wasn't able to reproduce reliably though.

Everything else appears to work correctly.

ghost:
@WithoutPants Thanks, I've fixed the performers search. I wasn't able to reproduce the state not updating on a rejected edit, I'm not sure how that would happen tbh, but there were a bunch of cache issues with merges and deletions, so I've added a force page reload after those.

Also finally fixed studio list ordering which was backwards.

PR #57, opened 2021-04-11
Fix incorrect table reference

peolic:
**This change was not tested.**

@InfiniteTF tagging you just in case it's not notifying you

Querying for user edits, like so:
```gql
query {
  queryEdits(edit_filter: { user_id: "<insert user id>" }) {
    count
    edits {
      status
    }
  }
}
```

Failed with error:
`Internal system error. Error <Error executing query: SELECT COUNT(*) as count FROM (SELECT edits.* FROM edits  WHERE scenes.user_id = $1) as temp, with args: [<user id>]: pq: missing FROM-clause entry for table "scenes">`

### DIFF
--- a/pkg/api/user_integration_test.go
+++ b/pkg/api/user_integration_test.go
@@ -469,6 +469,25 @@ func (s *userTestRunner) testRegenerateAPIKey() {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


ghost:
Thanks, works great.

PR #70, opened 2021-05-15
Move CI pipeline to Github actions

ghost:
Based the pipeline on https://github.com/stashapp/stash/pull/1295

I've moved the crosscompilation to goreng/golang-cross which provides golang 1.16. Since stash-box doesn't use cgo I've ditched most of the linker options in favor of the goreleaser defaults.

Once go-migrate cut a new release with support for 1.16 we can ditch packr as well in favor of golang embed, which should remove some additional clutter from the make process.


WithoutPants (approved):
The `meeDamian` action was causing bad tags when overwriting the development tag a little while ago, but this appears to now be fixed.

The docker credentials have been added to the repo.

.github/workflows/build.yml
```diff
@@ -0,0 +1,179 @@
...
+      with:
+        go-version: 1.16.x
+
+    - name: Install Node
+      uses: actions/setup-node@v2
+      with:
+        node-version: '14'
+
+    - name: Checkout
+      run: git fetch --prune --unshallow --tags
```
peolic, line 45: I noticed this on the stash repo too...
The docs for `actions/checkout` say [you can disable the shallow checkout](https://github.com/actions/checkout/blob/main/README.md#Fetch-all-history-for-all-tags-and-branches).
So theoretically you should be able to remove this step, and instead amend the first step to be:
```yml
    - uses: actions/checkout@v2
      with:
        fetch-depth: 0

PR #73, opened 2021-05-15
Fix issue when using `parentStudio` filter

peolic:
I forgot I made this commit...
~I still need to test this, but you can merge it if it's good.~
Tested and it does not fix the issue.

---

for example, when used in conjunction with performers filter, it fails due to invalid SQL query
(is missing a space after the body concatenation that is replaced in this commit)

For example query:
```gql
{
  queryScenes(
    scene_filter: {
      parentStudio: "<studio id>"
      performers: {
        modifier: INCLUDES
        value: ["<performer id>"]
      }
    }
  ) {
    count
    scenes {
      title
      performers {
        as
        performer {
          name
        }
      }
    }
  }
}
```

Error:
`Error executing query: SELECT COUNT(*) as count FROM (SELECT scenes.* FROM scenes LEFT JOIN studios ON scenes.studio_id = studios.idJOIN scene_performers ON scene_performers.scene_id = scenes.id WHERE (studios.parent_studio_id = $1 OR studios.id = $2) AND scene_performers.performer_id IN ($3) GROUP BY scenes.id ) as temp, with args: [<studio id> <studio id> <performer id>]: pq: syntax error at or near "scene_performers"`

### DIFF
--- a/pkg/database/sql.go
+++ b/pkg/database/sql.go
@@ -47,7 +47,7 @@ func NewDeleteQueryBuilder(table Table) *QueryBuilder {
 }
 
 func (qb *QueryBuilder) AddJoin(joinTable Table, on string) {
-	qb.Body += "JOIN " + joinTable.Name() + " ON " + on
+	qb.Body += " JOIN " + joinTable.Name() + " ON " + on
 	qb.Distinct = true
 }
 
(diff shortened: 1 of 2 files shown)


peolic:
This works now, though I went back and forth on how to best accomplish this fix.
At one point I added `joinClauses` to `QueryBuilder` and constructed them like where/group by/having in `buildBody`,
but then I realized that I removed the `LEFT` part of the `JOIN` clause,
so I reverted and did the simplest fix, which is to just add a space before each clause.

PR #84, opened 2021-05-23
Add linting and additional error checkers

ghost:


pkg/api/server.go
```diff
@@ -28,12 +28,12 @@ import (
 )
 
 var version string = "0.0.0"
```
peolic: `pkg/api/server.go:30:13: should omit type string from declaration of var version; it will be inferred from the right-hand side`
```suggestion
var version = "0.0.0"
```

peolic:
`make lint` should be added to the CI workflow too?

go.mod
```diff
@@ -5,65 +5,39 @@ go 1.16
...
 	github.com/rs/cors v1.6.0
-	github.com/russross/blackfriday/v2 v2.1.0 // indirect
 	github.com/sirupsen/logrus v1.8.1
-	github.com/spf13/cobra v1.1.3 // indirect
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.7.0
-	github.com/ugorji/go v1.1.4 // indirect
-	github.com/urfave/cli v1.20.0 // indirect
-	github.com/urfave/cli/v2 v2.3.0 // indirect
-	github.com/vektah/dataloaden v0.3.0 // indirect
```
peolic, line 58: I tried `go mod tidy` yesterday and it did exactly that.
This broke `make generate-dataloaders` for me (and maybe a few other things too, can't remember now).
Also tried to find a way to 'tell' `tidy` to keep this even if it's deemed unused (other than importing it somewhere in the code), but to no avail...

ghost: Guess I probably shouldn't have run tidy. In any case, readded the missing deps.

peolic: oddly enough, `make generate-dataloaders` still fails even after `go get` and `go mod vendor`
```
cannot find package "." in:
        /path/to/stash-box/vendor/github.com/vektah/dataloaden
```

ghost: ~~Did you go mod vendor after pulling again?~~ Duh.

peolic: running `go install` seems to have fixed it (possibly `go get github.com/vektah/dataloaden` outside the git repo too)

PR #88, opened 2021-05-24
Prevent creating body mod if it already exists

peolic:
* Trim body mod strings
* Simplify body mod creation


ghost:
Nice, I didn't know about `isValidNewOption`. Also good call on trim and feedback.

ghost (approved):

PR #91, opened 2021-05-28
Add studio filter to performer scene list

ghost:


frontend/src/components/checkboxSelect/CheckboxSelect.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </div>
+      ) : (
+        <div className="d-flex">
+          <Form.Check checked={unselected.includes(option.value)} />
+          <span className="text-muted">{option.label}</span>
+        </div>
+      );
+    return `${
+      unselected.length === 0 ? "All" : unselected.length
+    } studios selected`;
```
peolic:
```suggestion
    } ${plural} selected`;
```

frontend/src/graphql/queries/index.ts
```diff
@@ -81,6 +86,15 @@ export const usePerformer = (
     skip,
   });
 
+export const useFullPerformer = (
+  variables: PerformerVariables,
+  skip: boolean = false
```
peolic, line 91: This is never executed with `skip=true`, but you might be planning ahead for something.

ghost: That's just boilerplate, so it's possible to skip the query if necessary.

frontend/src/components/checkboxSelect/CheckboxSelect.tsx
ghost: Thanks, fixed.

PR #97, opened 2021-06-12
don't split scene performer names in the middle

peolic:
See `Monica Sweetheart` in the example below:

before
after

...and
before
after


peroo (approved):

peroo (changes requested):
.

PR #105, opened 2021-07-06
Fix changing description on an existing body mod location

peolic:
This fixes the issue I reported on Discord.
>it seems that adding a description to an existing piercing location gets ignored on performer merge (edit, too, probably)
for example, a performer had `Lip` with no description, I added two more piercings and a description to `Lip`, it __showed up on the edit preview__, but it didn't end up in the submitted edit details.

This is the most basic way of fixing it.
I believe this would actually remove and add, instead of updating the current,
but I don't know if it actually needs the extra logic to update the current location (requires changes to `PerformerRepo#ApplyModifyEdit`).


ghost:
Had to modify it a bit so it checks the string contents as well, but should be good now.

PR #115, opened 2021-08-18
User fingerprints

### LINKED ISSUE #99: [RFC] Fingerprint submission changes
# Problem

The current fingerprint submissions system is prone to bad data which involves manual labour to remove. This stems from the fingerprint `submissions` field being a simple counter which is incremented whenever a user matches it. Removing bad matches currently requires manually removing the entry.

# Proposal

Fingerprints submissions should be related to the user that submitted them, and a user can only pair a fingerprint hash to a single scene. That is, if they re-submit the same hash to a different scene, then the original user/fingerprint/scene relationship will be removed in favour of the new one. The `submissions` field will then be a resolver of a count of total submissions in the database. From what I can tell, the existing `submissions` field isn't actually maintained.

Querying for scenes by fingerprint may return multiple candidates, and we should allow for this in the graphql schema and in the client tagger view. Obviously, the best match (ie the one with the highest submission count) should be displayed first, but other candidates should be available in case the first is not the correct match. If a user currently has a scene matched, then they should be able to unmatch it manually, as an alternative to matching it to another scene (in the event that an alternative scene does not exist).

# Impacts

The one-scene-per-hash change will have impacts on the shared test account we currently have.

In order to maintain our existing fingerprint data, we will want to migrate it to link it with a specially created internal user.

# Other considerations

I don't think that it's necessary to have a report or down-vote et al function for scene fingerprints. I believe that bad matches will tend toward the zero count, and good matches will inevitably tend upwards. When a user finds a bad match, they can unmatch on there system if necessary, and manually look for an alternative scene and match that as needed.

To reduce useless or bad data, we should consider automating the removal of single count fingerprints that are older than a configurable age. This will reduce storage overhead.

I'm not sure what use `duration` will be when moving toward this new mechanism. We should be using the "official" duration for match validation in the client. Submitting clips or compilations from a single scene should also be encouraged, so the duration may not necessarily match. 

# Implementation details

## Database schema

In `scene_fingerprints`:
* deprecate `duration` and `submissions`
* remove `updated_at`
* add `user_id uuid`
* add indexes on `user_id`, `scene_id` and `created_at`

## Graphql schema

* add `unmatch` boolean flag to `FingerprintSubmission`. When true, this will unmatch the scene from the fingerprint for the current user, if present.
* deprecate `duration` from `Fingerprint` and related objects - make it optional.
* add `limit` and `offset` to `FingerprintQueryInput` to designate the maximum number of matches to return and the offset. This means that a client can get the top match(es) and alternative matches as needed.
* `findSceneByFingerprint` should return `QueryScenesResultType`, or some intermediate type to prevent breaking changes. The main purpose is so the client can get the total number of matching scenes for a single fingerprint.
* likewise, `findScenesBy[Full]Fingerprints` should return an array of `QueryScenesResultType`, or some intermediate type to prevent breaking changes.
* `fingerprints` should probably be deprecated from `Scene` (and return an empty list) in favour of a query interface. The number of fingerprints could potentially get quite large.

WithoutPants:
Resolves #99 

- Added new `user_id` column to `scene_fingerprints`. The migration creates an internal user `_legacy_submissions` and assigns the existing fingerprints to this user. All fingerprints will have their `submissions` counter reset to 1 as a result. This migration requires the `pgcrypto` extension to be installed to generate the created user id. 
- Added `user_ids` to `FingerprintInput`. This allows multiple users to be assigned to a single fingerprint when mutating scene fingerprints. If it is empty, or if the user does not have the modify role, then this is set to the current user only.
- Adds `unmatch` flag to `FingerprintSubmission`. When true, unmatches the fingerprint to the scene for the listed or current user
- `submissions` and `updated` has been deprecated and is now unused in `FingerprintEditInput`. Submissions is now derived from the number of user-hash relationships, and updated is derived this way as well.
- moved the user authorisation code into the `user` package
- refactored the scene integration tests (and some infrastructure code) to use a graphql client instead of poking into the resolvers manually
- moved scene mutation code into a new `scene` package

When getting scene fingerprints, the `submissions` field is resolved from the count of the scene/algorithm/hash group. `duration` is resolved as the average of the `duration` fields for the group. This is not ideal, but I'm not sure how else to handle it. `created` and `updated` are based on the `min` and `max` creation dates.

The update operation has been changed to maintain existing UI behaviour. This means that it removes any fingerprints that aren't in the provided input, maintains existing fingerprints that are provided (without changing the associated users), and adds any new fingerprints provided (setting the user id to the current user if not provided).


WithoutPants:
As part of this change, I've removed the ability to edit scene fingerprints with scene edits. Scene fingerprints should be manipulated using the `submitFingerprint` instead. 

I have also changed the output of `duration` to be the `mode` of the submissions, rather than the average. This means that the most popular duration value of the submissions for a hash will be shown.

ghost:
To support stash I think we need a field on Fingerprints to denote whether a user has submitted it or not. That way we can display that in stash and prompt the user to remove it if necessary. It would also make it a lot easier to manage fingerprint submissions after a scene has been tagged.

I also think it would be useful to have a `Submissions` field on fingerprints that detail the individual submissions, but that can wait for a separate PR.

WithoutPants:
> To support stash I think we need a field on Fingerprints to denote whether a user has submitted it or not. That way we can display that in stash and prompt the user to remove it if necessary. It would also make it a lot easier to manage fingerprint submissions after a scene has been tagged.

Done. I added a `user_submitted` field on `Fingerprint`. It required an extra query to match the user submitted fingerprints with the aggregate groups. 

I also changed the fingerprints view to show if the current user has submitted a fingerprint:

image

> I also think it would be useful to have a `Submissions` field on fingerprints that detail the individual submissions, but that can wait for a separate PR.

There is already a numeric `submissions` field on `Fingerprint`. I think I need more context as to what this would be used for.

PR #117, opened 2021-08-19
Add name filtering to users list

ghost:


frontend/src/pages/users/Users.tsx
```diff
@@ -62,6 +90,7 @@ const UsersComponent: React.FC = () => {
         setPage={setPage}
```
peolic, line 90: This is missing to make the pagination work correctly:
```suggestion
        setPage={setPage}
        perPage={PER_PAGE}
```

Issue #116, opened 2021-08-19
[RFC] Scene import process


WithoutPants:
# Problem

Need to provide a native data import process which handles mapping to other options. The process should allow for the creation of complete scene data without needing to perform post-import modifications.

# Proposal

This RFC is based on the previous work in the branch at https://github.com/InfiniteTF/stash-box/tree/bulk-import

Here is the basic requirements I think are needed for the mvp import functionality:
* accept csv (json to come later)
* support mapping of csv columns to scene fields
* support fixed scene fields (not just studio)
* support manipulation of csv data via regex replacements
* support mapping studio/performer/tag names to existing objects

Here is how I envision the import process to work:
* user navigates to the import page
* user selects a file to import, configures scene field to column mappings (and fixed values), and clicks submit
* system parses the file, inserting the data into a database table, and attempts to match studio/tag/performer names to objects
* system presents the results and allows the user to map/change studio/tag/performer names from the data to actual object instances (using auto-complete selectors)
* user maps studio/tag/performer names to objects
* user clicks to submit 
* system imports the data and removes the data from the import data table

# Other considerations
* only one import operation may be performed per user. User may resume or abort processing of pending imports.
* stale pending import data should be removed eventually

# Implementation details

## Database schema

Add new `import_data` table:
* `user_id` - assuming a single import per user, this plus `row` should be sufficient composite key
* `row` - index of csv or json
* `data` - a `jsonb` encoded of the raw row data

If we need to identify stale imports, then we'll need a separate table to identify pending imports, with a user and date.

## Graphql schema

* add `submitImport` mutation
  * accepts `data` file upload, a datafile `type`, and a list of `fieldMapping` objects
  * `ImportFieldMappingInput` includes `outputField`, optional `inputField`, optional `fixedValue`, optional list of regex replacements
  * processes the input file inserts into import table
* add `queryImportData` query, accepting page size and page. Returns parsed import data (does not translate tag/studio/performers).
* add `importMappings` query.
  * returns `ImportMapping` object, which contains `studios`, `performers`, `tags` fields, which are lists of name/id pairs
* `completeImport` mutation
  * accepts `ImportMappingInput` object which is basically the same structure as `ImportMapping`
  * creates scenes from the imported data, mapping tags/performers/studios as needed
  * names without a mapped id are ignored
* `abortImport` mutation
  * aborts a pending import for the current user, removing the data from the import data table

## GUI changes

* add `Import` top-level menu item for allowed users
[...]

PR #120, opened 2021-08-20
[WIP] Bulk scene import

WithoutPants:
Built off Infinite's branch and based on the design in #116.

- Added new user roles for submitting and finalising imports.
- Limited to a single pending import per user.
- Supports csv only
- Added import top-level menu that is only visible for authorised users

image

Allows configuration of scene field to input data file field or a fixed value (ie what "Main Studio" was apparently doing). Allows multiple per-field regex replacements, which are performed in order. Configurations can be saved and loaded to/from a json file.

Not pictured is the list delimiter field. For fields that allow multiple values (tags, performers), this string is used to delimit values. This is applied after the regex replacements.

After submitting (or if there is a pending import):

image

Allows browsing of the data to be imported. Not shown in this image, but where a studio/tag/performer is matched, the text is normal white and an applicable icon is shown. The scene images are also not shown in the above image, but are supported.

Pending imports can be completed or aborted. 

TODO:
- allowing changing of performer/tag/studio name to object mappings isn't added yet. There could potentially be a large number of these to deal with, so need to think about an appropriate way to deal with it
- importing scenes with an image URL is super slow and probably not done in a sustainable way - it currently just slurps all the URLs as needed. This probably needs to be offloaded to a separate task and throttled appropriately. An alternative is to accept encoded images in the data files and just use those instead of scraping them from the source.


ghost:
Thanks for picking this up!

Saving the import data in the db is a good idea. I would expand on that and allow multiple pending imports per user, so it's possible to for instance automatically upload datasets with preset configs for manual tweaking/approval. 

I also think it would help a lot if it was possible to modify the config after an import has been uploaded. Ideally it would just be a tabbed interface with the config in one pane, and the output in the other, so you could quickly modify regular expressions and see the updated results.

Regarding the main studio field that I added, the intention was that it could represent the network in the case where scenes might be from different sub studios of a scraped site. In your model we could accomplish the same thing with an optional parent studio field, then the parent studio could be fixed and the studio field dynamically set per scene.

It's no longer possible to parse durations from what I can tell. I used regexes with named capture groups, so you could do `(?P<hours>\d+)h, = (?P<minutes>\d+) minutes` for instance, to parse a particular duration format a site might use. I don't think replacement fields are sufficient for that case. The same goes for dates, I used the `dateparse` library for that.

ghost:
> * importing scenes with an image URL is super slow and probably not done in a sustainable way - it currently just slurps all the URLs as needed. This probably needs to be offloaded to a separate task and throttled appropriately. An alternative is to accept encoded images in the data files and just use those instead of scraping them from the source.

The latter is what I ended up doing when scraping. Throw all images in an async queue and let them finish on their own accord, otherwise it holds up the process too much.

EDIT: The former I mean. Uploading images with the data would work, although it complicates the process a bit.

pkg/manager/bulkimport/csv.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"encoding/csv"
+	"io"
+)
+
+type csvParser struct{}
+
+func (csvParser) parse(input io.Reader, fn processFn) error {
+	reader := csv.NewReader(input)
```
bnkai: For importing large csv files setting `ReuseRecord` reduces the allocations a lot. It shouldnt cause any trouble since every row seems to be processed and then discarded anyway.
```golang
	reader := csv.NewReader(input)
  csvreader.ReuseRecord = true
```

WithoutPants:
> Regarding the main studio field that I added, the intention was that it could represent the network in the case where scenes might be from different sub studios of a scraped site. In your model we could accomplish the same thing with an optional parent studio field, then the parent studio could be fixed and the studio field dynamically set per scene.

We're only setting a single studio on a scene. Do you mean a default studio if no specific studio is set? Adding a default field value shouldn't be too difficult, if I understand correctly.

> I also think it would help a lot if it was possible to modify the config after an import has been uploaded. Ideally it would just be a tabbed interface with the config in one pane, and the output in the other, so you could quickly modify regular expressions and see the updated results.

I'll have to reconsider the upload process. At the moment, it requires the config while uploading so that we can eliminate unnecessary data. Instead, the process will have to be to upload the file first, then select the columns and options, with the ability to preview the results.

ghost:
> > Regarding the main studio field that I added, the intention was that it could represent the network in the case where scenes might be from different sub studios of a scraped site. In your model we could accomplish the same thing with an optional parent studio field, then the parent studio could be fixed and the studio field dynamically set per scene.
> 
> We're only setting a single studio on a scene. Do you mean a default studio if no specific studio is set? Adding a default field value shouldn't be too difficult, if I understand correctly.

I mean when creating new studios. For instance when importing a report from studio A, which has sub-studios B,C,D, etc denoted in the column `studios`. In this case you would set the `parent studio` field to A so that if a new studio is encountered it would get the correct parent.

> > I also think it would help a lot if it was possible to modify the config after an import has been uploaded. Ideally it would just be a tabbed interface with the config in one pane, and the output in the other, so you could quickly modify regular expressions and see the updated results.
> 
> I'll have to reconsider the upload process. At the moment, it requires the config while uploading so that we can eliminate unnecessary data. Instead, the process will have to be to upload the file first, then select the columns and options, with the ability to preview the results.

Right, I didn't consider that we have to store the raw report. I think that should be fine though. An advantage of uploading first is that you can sniff the column names and populate title/performers/duration, etc. in the column options immediately.

WithoutPants:
> I mean when creating new studios. For instance when importing a report from studio A, which has sub-studios B,C,D, etc denoted in the column `studios`. In this case you would set the `parent studio` field to A so that if a new studio is encountered it would get the correct parent.

Rightio. The current iteration doesn't handle creating studios/performers/tags as part of the scene import process, so I might defer that to a later iteration.

WithoutPants:
Changed the process to be three-step: upload file, massage the data, then map and complete the import. 

The massage view shows the field config like in the first screenshot, plus a view of the data rows. Shows matched and unmatched fields:
image

Moved list delimiter to be per-field.

Import process is still single-threaded and very slow, but it now logs the scenes as they are imported.

Makefile
```diff
@@ -88,6 +88,14 @@ pre-ui:
 ui:
 	cd frontend && yarn build
 
+.PHONY: ui-fmt
+ui-fmt:
+	cd frontend && yarn format
+
+.PHONY: ui-start
+ui-start: pre-build
```
peolic: Did you mean?
```suggestion
ui-start: pre-ui
```

WithoutPants: No. `pre-build` is used to populate the build environment variables. In stash these are passed into the react compile process to embed version information. It is not currently done here, but it doesn't hurt that its there.

WithoutPants:
I think this really needs a job management system to work well. I'm going to defer this and add job management, so that the import system can use it.

PR #122, opened 2021-08-23
Fix body mods

peolic:
Fixes an issue where the body mods were not getting updated.
Similar to the `EditImages` component.
Had to deal with stripping `__typename` from the data, not sure if there is a better way to do that.

### DIFF
--- a/frontend/src/components/form/BodyModification.tsx
+++ b/frontend/src/components/form/BodyModification.tsx
@@ -1,18 +1,24 @@
-import React, { useState, useRef } from "react";
+import React from "react";
 import Creatable from "react-select/creatable";
 import { components } from "react-select";
 import { Button, Form, InputGroup } from "react-bootstrap";
-import { Controller } from "react-hook-form";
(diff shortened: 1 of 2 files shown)


frontend/src/components/form/BodyModification.tsx
```diff
@@ -1,76 +1,76 @@
...
+            Location
+          </InputGroup.Text>
+        </InputGroup.Prepend>
+        <Form.Control defaultValue={mod.location} readOnly />
+        <Form.Control
+          defaultValue={mod.description ?? ""}
+          placeholder={descriptionPlaceholder}
+          onInput={(e: React.ChangeEvent<HTMLInputElement>) =>
+            update(index, { ...mod, description: e.currentTarget.value })
+          }
```
peolic, line 64: There's probably a better way to do that.

ghost: You should be able to use a `<Controller>` as well, but this is very straightforward so no complaints from me.

PR #129, opened 2021-09-08
Add voting on edits

ghost:
Adds voting on edits with automatic application of edits after three unanimus positive votes (and 48h in the case of destructive edits), or after four days if the threshold hasn't been reached.

Also adds automatic user voting rights after ten successful edits, to prevent vandalism and manual management of rights.


WithoutPants:
Notes after spending some time testing:
* users are able to vote on their own edits, which seems wrong to me
* the current vote display in the UI is broken. I would expect it to be the net sum of all positive and negative votes. Instead, it looks like it takes the sign of the current net total, then appends the total vote amount. ie 3 positive, 1 negative results in +4. Seems to be further broken when abstain is used
* is it necessary to show abstain votes from users?
* I was able to trigger an accepted application when hitting the +3 threshold, but could not trigger a rejection, even with 4 no votes

WithoutPants:
Also it looks like the cronjob might be a little screwy. I got the following log errors while running:
```
Edit cronjob failed to apply edit 0a11eef3-1a27-49dc-ab8c-02682c2de5cc: user not found: a7d1e8e2-1c17-41ce-a2db-07fd3cfb30ed
```

Said edit was one that had already been accepted.

peolic:
Edits that have already been submitted by the time this functionality goes live,
are (most of them, anyway) already past the voting period and may get applied before any votes can be cast.
Just wanted to point that out.
Obviously, there are ways to mitigate that even without code changes, but maybe it's worth some temporary code change like only applying the voting period only after a specific timestamp.

ghost:
> * users are able to vote on their own edits, which seems wrong to me

Fixed
> * the current vote display in the UI is broken. I would expect it to be the net sum of all positive and negative votes. Instead, it looks like it takes the sign of the current net total, then appends the total vote amount. ie 3 positive, 1 negative results in +4. Seems to be further broken when abstain is used

Whoops, fixed.
> * is it necessary to show abstain votes from users?

Not really, I've removed it.
> * I was able to trigger an accepted application when hitting the +3 threshold, but could not trigger a rejection, even with 4 no votes

I forgot to add that. Works same as `{threshold}` positive votes now.

> Also it looks like the cronjob might be a little screwy. I got the following log errors while running:

Should be fixed now. The user promotion didn't handle deleted/nonexistant users correctly.

> Edits that have already been submitted by the time this functionality goes live,
are (most of them, anyway) already past the voting period and may get applied before any votes can be cast.

Good point, I've added a step to the migration to reject all pending edits to avoid stale edits getting auto applied. I'll make sure to clean out the edit queue in stashdb before pushing live.

pkg/manager/cron/cron.go
```diff
@@ -0,0 +1,73 @@
...
+			applyEdit := false
+			if e.Operation == models.OperationEnumDestroy.String() || e.Operation == models.OperationEnumMerge.String() {
+				if e.VoteCount > 0 {
+					applyEdit = true
+				}
+			} else {
+				if e.VoteCount >= 0 {
+					applyEdit = true
+				}
+			}
```
WithoutPants: What's going on here?

pkg/sqlx/querybuilder_edit.go
```diff
@@ -314,3 +333,20 @@ func (qb *editQueryBuilder) FindByStudioID(id uuid.UUID) ([]*models.Edit, error)
 func (qb *editQueryBuilder) FindBySceneID(id uuid.UUID) ([]*models.Edit, error) {
 	return qb.findByJoin(id, editSceneTable, "scene_id")
 }
+
+func (qb *editQueryBuilder) FindCompletedEdits(votingPeriod int, minimumVotingPeriod int, minimumVotes int) ([]*models.Edit, error) {
```
WithoutPants, line 341: This needs documentation as to what this actually does. I'm not able to fully work it out looking at the code alone.

WithoutPants:
- Three no votes causes the edit to be accepted.

I think there needs to be an indication in the GUI of the deadline before a pending edit will be accepted or rejected, since currently that aspect of the system is a bit opaque from the user's point of view.

pkg/manager/config/config.go
```diff
@@ -39,6 +39,17 @@ type config struct {
 	EmailCooldown     int      `mapstructure:"email_cooldown"`
 	DefaultUserRoles  []string `mapstructure:"default_user_roles"`
 
+	// Number of approved edits before user automatically gets VOTE role
+	VotePromotionThreshold int `mapstructure:"vote_promotion_threshold"`
+	// Number of positive votes required for immediate approval
+	VoteApplicationThreshold int `mapstructure:"vote_application_threshold"`
+	// Duration, in seconds, of the voting period
```
WithoutPants, line 46: This needs to be expanded further. What is the "voting period"?

pkg/manager/cron/cron.go
ghost: When closing edits it's differentiating between destructive and non-destructive edits. Destructive edits require at least +1 votes to pass. Non-destructive pass even with no votes or tie votes since they're easier to revert.

WithoutPants:
I also think that this PR is a good time to add a comprehensive document detailing how the voting system works. This will form the basis of any testing that we do.

pkg/manager/cron/cron.go
WithoutPants: For clarity, can this be rewritten to something like:
```
voteThreshold := 0
if e.Operation == models.OperationEnumDestroy.String() || e.Operation == models.OperationEnumMerge.String() {
	// require at least +1 votes to pass destructive edits
	voteThreshold = 1
}

...
if e.VoteCount >= voteThreshold {
	_, err = edit.ApplyEdit(c.rfp.Repo(), e.ID, false)
} else {
	_, err = edit.RejectEdit(c.rfp.Repo(), e.ID, false)
}
```

pkg/manager/edit/edit.go
```diff
@@ -49,26 +52,43 @@ type editApplyer interface {
...
 }
 
-func ApplyEdit(fac models.Repo, editID uuid.UUID) (*models.Edit, error) {
+func validateEditPresence(edit *models.Edit) error {
+	if edit == nil {
+		return errors.New("Edit not found")
+	}
+
+	if edit.Applied {
+		return errors.New("Edit already applied")
```
WithoutPants: Lowercase the error strings.

pkg/manager/edit/edit.go
```diff
@@ -151,3 +207,40 @@ func urlCompare(subject []*models.URL, against []*models.URL) (added []*models.U
...
+	if err != nil {
+		return models.VoteStatusEnumPending, err
+	}
+
+	positive := 0
+	negative := 0
+	for _, vote := range votes {
+		if vote.Vote == string(models.VoteTypeEnumAccept) {
+			positive++
+		} else if vote.Vote == string(models.VoteTypeEnumReject) {
```
WithoutPants: Use `String()` instead.

pkg/models/user.go
```diff
@@ -17,6 +17,9 @@ type UserRepo interface {
 	Count() (int, error)
 	Query(userFilter *UserFilterType, findFilter *QuerySpec) (Users, int)
 	GetRoles(id uuid.UUID) (UserRoles, error)
+	CountSuccessfulEdits(id uuid.UUID) (int, error)
+	CountFailedEdits(id uuid.UUID) (int, error)
```
WithoutPants: These should probably be renamed to `CountAcceptedEdits` and `CountRejectedEdits` since we also have the concept of failed edits, which the latter does not seem to apply to. Incidentally, the fail enum value ideally should be documented to highlight the difference between it and `Rejected`.

pkg/api/resolver_model_user.go
```diff
@@ -24,24 +24,28 @@ func (r *userResolver) Roles(ctx context.Context, user *models.User) ([]models.R
...
+	return qb.CountVotesByType(obj.ID, models.VoteTypeEnumAccept)
 }
 
-func (r *userResolver) UnsuccessfulVotes(ctx context.Context, user *models.User) (int, error) {
-	// TODO
-	return 0, nil
+func (r *userResolver) UnsuccessfulVotes(ctx context.Context, obj *models.User) (int, error) {
+	fac := r.getRepoFactory(ctx)
+	qb := fac.User()
+	return qb.CountVotesByType(obj.ID, models.VoteTypeEnumReject)
```
WithoutPants: These do not match what is documented in the graphql schema. From what I can tell from the code, these functions return the number of accept and reject votes respectively that they have made. The documented meaning of these fields are the number of (presumably accept) votes on successful and unsuccessful edits. These fields should probably be interrogated further as to their usefulness. I think the original intention was to show how useful a user appears in terms of their votes on ultimately successful or unsuccessful edits.

pkg/user/user.go
```diff
@@ -463,3 +463,42 @@ func getDefaultUserRoles() []models.RoleEnum {
 
 	return ret
 }
+
+func PromoteUserVoteRights(fac models.Repo, userID uuid.UUID, threshold int) error {
```
WithoutPants, line 467: Needs documentation.

WithoutPants (changes requested):

pkg/manager/cron/cron.go
ghost: That's a good suggestion, thanks!

WithoutPants (approved):
Looks ok from what I can see. Happy for this to go in as is and we'll look at adding the voting documentation later.

PR #138, opened 2021-10-08
Misc fixes

peolic:
- Fix missing scene/studio buttons


WithoutPants:
- Fixed an error that was preventing UI from compiling.
- Fixed up the styling on the image edit that caused the image to be proportioned incorrectly
- Changed behaviour of edits such that the next/submit button is disabled if an image is selected but not uploaded, and a status message is shown indicating this.
- Verified that create/edit buttons are shown for correct users.

peolic:
Everything you added works for me.

Added:
- Strike-through scene title when it is marked as deleted.
- Hide deleted scenes from scene lists and search results.

Another styling issue on edit cards: (the logo looks okay on the studio page)

image

PR #145, opened 2021-10-26
Migrate from CRA to Vite

ghost:
Props to @kermieisinthehouse for getting the ball rolling.

Needs more thorough testing, but so far no serious issues.


frontend/package.json
```diff
@@ -4,10 +4,8 @@
   "description": "Stash-box",
   "license": "MIT",
   "scripts": {
-    "start": "react-scripts start",
-    "build": "craco build",
-    "test": "react-scripts test",
-    "eject": "react-scripts eject",
+    "start": "vite",
+    "build": "vite build",
     "validate": "yarn lint && yarn format-check",
```
kermieisinthehouse: Vite doesn't typecheck Typescript, you need to add `&& tsc` either here before the build step

frontend/package.json
```diff
@@ -57,10 +50,12 @@
     "@fortawesome/react-fontawesome": "^0.1.15",
     "@hookform/resolvers": "2.8.1",
     "@types/apollo-upload-client": "^14.1.0",
+    "@vitejs/plugin-react": "^1.0.6",
```
kermieisinthehouse: You aren't using this

frontend/tsconfig.json
```diff
@@ -19,6 +19,7 @@
     "skipLibCheck": true,
     "strict": true,
     "strictFunctionTypes": true,
+    "types": ["vite/client"],
```
kermieisinthehouse: I found that this caused problems but some resources online said it's necessary, it might be for this codebase but if you have type errors, try without

frontend/package.json
ghost: Thanks, I've added it. We definitely want fast refresh.

frontend/tsconfig.json
ghost: Seems like it's not really useful, so no problem removing it.

PR #163, opened 2021-11-07
UI fixes

peolic:
Few things I noticed.
Feel free to add other commits.

### DIFF
--- a/frontend/src/components/changeRow/ChangeRow.tsx
+++ b/frontend/src/components/changeRow/ChangeRow.tsx
@@ -16,7 +16,7 @@ const ChangeRow: FC<ChangeRowProps> = ({
 }) =>
   name && (newValue || oldValue) ? (
     <div className="row mb-2">
-      <b className="col-2 text-end">{name}</b>
+      <b className="col-2 text-end pt-1">{name}</b>
       {showDiff && (
         <span className="col-5">
           <div className="EditDiff bg-danger">{oldValue}</div>
--- a/frontend/src/components/editCard/MergeEdit.tsx
+++ b/frontend/src/components/editCard/MergeEdit.tsx
@@ -1,6 +1,6 @@
 import { FC } from "react";
 import { Link } from "react-router-dom";
-import { Row } from "react-bootstrap";
+import { Col, Row } from "react-bootstrap";
 import { faCheck, faTimes } from "@fortawesome/free-solid-svg-icons";
 
 import {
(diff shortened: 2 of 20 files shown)


peolic:
the `VoteBar` issue still remains, not sure how to fix it..

Bootstrap 4|Bootstrap 5
--|--
image|image

peolic:
when voting on an edit that can't be applied due to a conflict,
if the new vote count is past the threshold there's a (hidden) error in the browser console and the "Save" button does not update (the vote is registered in the database, though).
and the edit does not fail like it should.
```
[GraphQL error]: Message: Invalid addition. ID already exists '<alias>', Location: undefined, Path: editVote
Uncaught (in promise) Error: Invalid addition. ID already exists '<alias>'
```

less of a UI issue, but same goes for the cronjob, it does not fail conflicted edits when the voting period is over,
as seen on the staging instance
`ERRO[0299] Edit cronjob failed to apply edit <uuid>: Invalid addition. ID already exists '<alias>'`

peolic:
Tag category seems to be mandatory (on `TagForm.tsx`), unsure if this was [an intentional change](https://github.com/stashapp/stash-box/commit/219e0522bcb15a6bd4008d8fa621a323078e42cf#diff-e1086ae132298a775168258fc2bff77d2842731ae8935552ee6bb28b887d4c51L24).
_Edit_: Pushed a commit to correct this for now. If this was intended, it needs a better error message.

ghost:
> when voting on an edit that can't be applied due to a conflict, if the new vote count is past the threshold there's a (hidden) error in the browser console and the "Save" button does not update (the vote is registered in the database, though). and the edit does not fail like it should.
> 
> ```
> [GraphQL error]: Message: Invalid addition. ID already exists '<alias>', Location: undefined, Path: editVote
> Uncaught (in promise) Error: Invalid addition. ID already exists '<alias>'
> ```
> 
> less of a UI issue, but same goes for the cronjob, it does not fail conflicted edits when the voting period is over, as seen on the staging instance `ERRO[0299] Edit cronjob failed to apply edit <uuid>: Invalid addition. ID already exists '<alias>'`

Edit failures are on my list of things to fix. It shouldn't return an error since it's not a request error.

> Tag category seems to be mandatory (on TagForm.tsx), unsure if this was an intentional change.

Not intentional as far as I can remember.

@peolic Got any more UI fixes or should I merge it?

peolic:
yeah looks good
just had this weirdness to fix, I believe I'm out of fixes now

Stf2cCWPw9

PR #171, opened 2021-11-17
Fix homepage scene cards columns

peolic:
When less than* some threshold (4 per row, maybe?), the columns are stretched to the max.
This tweak seems to satisfy both the low count and the high count of scenes correctly.

### Before:
before

### After (below threshold):
after

### After (max width on 1080p resolution):
after-max


ghost:
Good catch!

PR #172, opened 2021-11-17
Ensure order for studios

peolic:
Fixes unstable order of paginated results.


peolic:
Solution needs to be rethinked, as per Discord.

PR #173, opened 2021-11-17
UI fixes

peolic:
- Popover content had light-colored text over a light background (see `<Help />` on performer form when changing the name).
- The studio names on the studio selector (performer page) needed to be nudged slightly away from the check-boxes.
- Grouped together the rest of the default SCSS variables overrides.
- Set the default value of the studio/tag lists search box from query.
- Fix missing question marks in `formatMeasurements`.
- Fix styling of entity delete pages (buttons stretched across the width of the page).
- Fix missing close button on modals.
- Restore sourcemaps on production builds (only TS/JS. seems CSS is not supported ATM).
- Fix warning on `ImageChangeRow`
- Fix pagination links focus colors
- Always mount the scene fingerprints tab
- Fix hidden performer scenes filters (the new sort order select was hidden behind the studio select)
- Use `SceneList` on tag page

### DIFF
--- a/frontend/src/pages/performers/PerformerDelete.tsx
+++ b/frontend/src/pages/performers/PerformerDelete.tsx
@@ -50,42 +50,37 @@ const PerformerDelete: FC<Props> = ({ performer }) => {
     });
 
   return (
-    <>
-      <Form
-        className="PerformerDeleteForm"
-        onSubmit={handleSubmit(handleDelete)}
-      >
-        <Row>
-          <h4>
-            Delete performer <em>{performer.name}</em>
-          </h4>
-        </Row>
-        <Form.Control type="hidden" value={performer.id} {...register("id")} />
-        <Row className="my-4">
-          <Col md={6}>
-            <EditNote register={register} error={errors.note} />
-          </Col>
-        </Row>
-        <Row className="mt-2">
-          <Button
-            variant="danger"
-            className="ms-auto me-2"
-            onClick={() => history.goBack()}
-          >
-            Cancel
-          </Button>
-          <Button
-            type="submit"
-            disabled
-            className="d-none"
-            aria-hidden="true"
-          />
-          <Button type="submit" disabled={deleting}>
-            Submit Edit
-          </Button>
-        </Row>
-      </Form>
-    </>
+    <Form className="PerformerDeleteForm" onSubmit={handleSubmit(handleDelete)}>
+      <Row>
+        <h4>
+          Delete performer <em>{performer.name}</em>
+        </h4>
+      </Row>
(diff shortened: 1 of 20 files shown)


frontend/src/pages/performers/PerformerDelete.tsx
```diff
@@ -50,42 +50,37 @@ const PerformerDelete: FC<Props> = ({ performer }) => {
...
+              onClick={() => history.goBack()}
+            >
+              Cancel
+            </Button>
+            <Button
+              type="submit"
+              disabled
+              className="d-none"
+              aria-hidden="true"
+            />
```
peolic, line 76: This exists on all `*Delete` components. The actual submit button is below.
Can be removed? 🤔

ghost: It's in all forms I think. The purpose was to prevent pressing enter from submitting the form. Since the disabled submit button is first it gets the keypress event, and silently does nothing. I can't reproduce the issue after removing them now though. At the time pressing enter in an event note would submit, but that doesn't seem to be the case, so maybe it was fixed in the form library? In either case, if you can't reproduce the issue then removing should be fine.

peolic: Ah, I see.
Well, it doesn't happen on the delete forms solely because pressing enter on a `textarea`, being multiline, doesn't submit the form.
But it does still happen if you click enter on an input field on, say, performer form.

peolic:
sorting by "trending" on a performer/tag page generates a query error, requiring `GROUP BY T.count` to be added.
```
Error executing query:
	SELECT scenes.* FROM scenes
	JOIN scene_performers
		ON scene_performers.scene_id = scenes.id
	JOIN (
		SELECT scene_id, COUNT(*) AS count
		FROM scene_fingerprints
		WHERE created_at >= (now()::DATE - 7)
		GROUP BY scene_id
	) T ON scenes.id = T.scene_id
	WHERE scenes.deleted = ? AND scene_performers.performer_id IN (?)
	GROUP BY scenes.id
	ORDER BY T.count DESC
	LIMIT 40 OFFSET 0 ,
with args: [false 31d8907d-2495-4231-9211-b9ea51501083]:
pq: column "t.count" must appear in the GROUP BY clause or be used in an aggregate function
```
this is a hacky fix: https://github.com/peolic/stash-box/commit/10c5bc4b61f596dbb938b4560174d9e8513d3c5c

ghost:
> sorting by "trending" on a performer/tag page generates a query error, requiring `GROUP BY T.count` to be added.

I've refactor the distinct filtering in #176, I think that should cover it.

PR #177, opened 2021-11-22
Update performer search query

### LINKED ISSUE #124: [Feature] Consider performer aliases when searching
[Currently, only performer names are considered when searching for performers.](https://github.com/stashapp/stash-box/blob/363076d006b5928700f8f3340c04eecc86b096e2/pkg/sqlx/querybuilder_performer.go#L524-L534)
The search term should be checked against the performer's aliases too.

Maybe the performer disambiguation should be considered as well, although they might need to be given less weight, compared to the name and aliases.

peolic:
~~This is a start, based on the current full text search for scenes.~~

~~In regards to performance, I'm unsure:
Should there be a `performer_search` table and functions updating the values, like `scene_search`?~~

Closes #124

This adds disambiguation and aliases to the performer search query with lower weights,
enabling searching for performers using those fields.

### DIFF
--- a/pkg/sqlx/querybuilder_performer.go
+++ b/pkg/sqlx/querybuilder_performer.go
@@ -532,12 +532,19 @@ func (qb *performerQueryBuilder) GetAllPiercings(ids []uuid.UUID) ([][]*models.B
 
 func (qb *performerQueryBuilder) SearchPerformers(term string, limit int) (models.Performers, error) {
 	query := `
-        SELECT * FROM performers
-        WHERE name % $1
-        AND similarity(name, $1) > 0.5
-        AND deleted = FALSE
-        ORDER BY similarity(name, $1) DESC
-        LIMIT $2`
+		SELECT P.* FROM performers P
+		LEFT JOIN (
+			SELECT performer_id, string_agg(alias, ' ') aliases
+			FROM performer_aliases
+			GROUP BY performer_id
+		) PA on PA.performer_id = P.id
+		WHERE (
+			to_tsvector('english', coalesce(P.name, '')) ||
+			to_tsvector('english', coalesce(P.disambiguation, '')) ||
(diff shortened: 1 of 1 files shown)


peolic:
Also, if I'm entirely wrong and this should not be a full text search, please do let me know.

ghost:
The existing performer search uses trigrams so it can support partial matches and typos. FTS on the other hand is exact so typos will exclude results. It might be fine, but we would need to add weights to prioritize names over aliases and disambiguation.

FWIW I have a branch where I started wiring stash-box up to Elasticsearch. I shelved it to focus on shipping, but long-term that's likely going to be the solution rather than postgres search.

peolic:
I looked at `setweight` before submitting, but while trying it I couldn't tell if it was actually doing anything.
(IIRC, I tried 'A' for name, 'B' for aliases and 'C' for disambiguation)

peolic:
This also works if we want to stick to trigrams until elastic.
```golang
func (qb *performerQueryBuilder) SearchPerformers(term string, limit int) (models.Performers, error) {
	query := `
		SELECT DISTINCT ON (T.id) T.* FROM (
			SELECT P.* FROM performers P
			LEFT JOIN performer_aliases PA on PA.performer_id = P.id
			WHERE
				P.deleted = FALSE AND (
					(P.name % $1 AND similarity(P.name, $1) > 0.5)
					OR (PA.alias % $1 AND similarity(PA.alias, $1) > 0.6)
					OR (P.disambiguation % $1 AND similarity(P.disambiguation, $1) > 0.7)
				)
			ORDER BY
				similarity(P.name, $1) DESC,
				similarity(PA.alias, $1) DESC,
				similarity(P.disambiguation, $1) DESC
		) T LIMIT $2`
	args := []interface{}{term, limit}
	return qb.queryPerformers(query, args)
}
```

PR #193, opened 2021-11-27
Add failing of edits with edit notes

ghost:
* Refactors the performer edit handling to move logic out of the query builder, similar to scene handling.
* Add `StashBot` user which posts edit notes for failed edits.
* Fails edit when an error is encountered during application.
* Changes system user management to check for root/stashbot on boot every time.
* Adds tooltip for edit status with explanations for failing statuses.
* Removes fail condition when trying to remove a join that no longer exists, or add one that is already there. They are instead ignored.
* Adds `BOT` user role. Not used for anything right now, but might be useful in the future.

### DIFF
--- a/pkg/models/translate.go
+++ b/pkg/models/translate.go
@@ -15,7 +15,15 @@ type ErrEditPrerequisiteFailed struct {
 }
 
 func (e *ErrEditPrerequisiteFailed) Error() string {
-	return fmt.Sprintf("Expected %s to be “**%v**”, but was “**%v**”.", e.field, e.expected, e.actual)
+	expected := "_blank_"
+	if e.expected != "" {
+		expected = fmt.Sprintf("“**%s**”", e.expected)
+	}
+	actual := "_blank_"
+	if e.actual != "" {
+		actual = fmt.Sprintf("“**%s**”", e.actual)
+	}
+	return fmt.Sprintf("Expected %s to be %s, but was %s.", e.field, expected, actual)
 }
 
 // fromEdit translates edit object fields into entity fields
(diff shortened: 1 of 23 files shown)


pkg/models/translate.go
```diff
@@ -8,6 +8,16 @@ import (
...
 )
 
+type ErrEditPrerequisiteFailed struct {
+	field    string
+	expected interface{}
+	actual   interface{}
+}
+
+func (e *ErrEditPrerequisiteFailed) Error() string {
+	return fmt.Sprintf("Expected %s to be “**%v**”, but was “**%v**”.", e.field, e.expected, e.actual)
```
peolic: > Edit application failed:
>
> Prerequisite failed: Expected Details to be “text text text”, but was “****”.

Empty values cause the 'bold' markdown to not render.

ghost: Thanks, fixed!

pkg/user/user.go
```diff
@@ -502,3 +536,16 @@ func PromoteUserVoteRights(fac models.Repo, userID uuid.UUID, threshold int) err
 
 	return nil
 }
+
+func GetModUser(fac models.Repo) *models.User {
+	if modUser == nil {
+		user, err := fac.User().FindByName("StashBot")
```
peolic:
```suggestion
		user, err := fac.User().FindByName(modUserName)
```

peolic (approved):

peolic:
Also run UI generate. `RoleEnum` on `globalTypes.ts` is missing the new `BOT` role.

PR #200, opened 2021-11-28
Add gql directives for modify/vote/edit/read

ghost:
Makes permission handling a lot simpler since it's largely defined in the schema.

The downside is that the way we've been testing auth no longer works. We call the resolvers directly, while directives run ahead of resolvers. We should ideally change our tests to post actual graphql queries rather than calling resolvers. That said, the permission handling is very simple so I'm not particularly worried.


peolic:
There's also this: https://gqlgen.com/reference/directives/
```gql
directive @hasRole(role: RoleEnum!) on FIELD_DEFINITION
```

peolic (approved):
Looks OK to me.

PR #202, opened 2021-11-29
TagEdit category resolver

peolic:
This just seems better than calling the API.


pkg/api/resolver_model_tag_edit.go
```diff
@@ -0,0 +1,20 @@
...
+type tagEditResolver struct{ *Resolver }
+
+func (r *tagEditResolver) Category(ctx context.Context, obj *models.TagEdit) (*models.TagCategory, error) {
+	if obj.CategoryID == nil {
+		return nil, nil
+	}
+
+	qb := r.getRepoFactory(ctx).TagCategory()
+	categoryID, _ := uuid.FromString(*obj.CategoryID)
+	return qb.Find(categoryID)
```
peolic: If this PR is merged before #199, this needs to be modified.
```suggestion
	return qb.Find(*obj.CategoryID)
```

ghost: I recommend using a dataloader for this seeing as a scene could have a lot of tags, most of which will have the same categories. The tagResolver already uses one for categories.

ghost: Duh, no, this is a tagEdit, nevermind.

ghost (approved):
Yup, definitely better than using graphql.

PR #208, opened 2021-11-30
Optimize pending edits query

peolic:
Only query the count. Probably marginal but the entire edits are just not needed for this.


frontend/src/graphql/queries/PendingEdits.gql
```diff
@@ -0,0 +1,8 @@
+query PendingEdits($type: TargetTypeEnum!, $id: ID!) {
```
peolic: Maybe should be named `PendingEditsCount`, or some other thing?

ghost: Yeah, `PendingEditsCount` sounds fine to me.

ghost (approved):

PR #214, opened 2021-12-03
Fix initial root password

peolic:
The password of the bot user was being printed.


ghost (approved):
Duh, I probably should've tested that first. Thanks!

PR #216, opened 2021-12-05
Fix scenes fingerprints criterion

peolic:
`/scenes?fingerprint={hash}` is currently broken:
`[GraphQL error]: Message: uuid: incorrect UUID length 16 in string "<hash>", Location: undefined, Path: queryScenes,scene_filter,fingerprints,value,0`

~~Not sure how to proceed from this point, yet.~~

I assume the correct fix is an interface that makes the two `Multi` types compatible with each other so they can be used in a single `getMultiCriterionClause` function, ~but that's where I got stuck.~


ghost (approved):
Nice solution!

PR #217, opened 2021-12-07
Fix edit expiration

peolic:
- I noticed the UI did not display the correct expiration time for destructive edits.
Making it a field makes it easy to update the conditions that label an edit as destructive.
(I was considering labelling performer renames as destructive when I noticed the issue)

- `ScenesList` sort order selector was wider than needed.

### DIFF
--- a/frontend/src/components/editCard/VoteBar.tsx
+++ b/frontend/src/components/editCard/VoteBar.tsx
@@ -26,12 +26,22 @@ const VoteBar: FC<Props> = ({ edit }) => {
   const [vote, setVote] = useState<VoteTypeEnum | null>(userVote?.vote ?? null);
   const [submitVote, { loading: savingVote }] = useVote();
 
-  if (
-    edit.status !== VoteStatusEnum.PENDING ||
-    !canVote(auth.user) ||
-    auth.user?.id === edit.user?.id
-  )
+  if (edit.status !== VoteStatusEnum.PENDING || !canVote(auth.user))
     return <></>;
+
+  if (auth.user?.id === edit.user?.id)
+    return (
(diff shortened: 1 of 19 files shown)


peolic:
On the topic of edit voting, I raised this on Discord not too long ago:
Should the user who submitted an edit be able to see the current vote score (just the score, without the buttons) on their own edit?

frontend/src/components/editCard/VoteBar.tsx
```diff
@@ -26,13 +26,23 @@ const VoteBar: FC<Props> = ({ edit }) => {
...
-    edit.status !== VoteStatusEnum.PENDING ||
-    !canVote(auth.user) ||
-    auth.user?.id === edit.user?.id
-  )
+  if (edit.status !== VoteStatusEnum.PENDING || !canVote(auth.user))
     return <></>;
 
+  if (auth.user?.id === edit.user?.id)
+    return (
+      <div className={CLASSNAME}>
```
peolic: I think it looked bad without the buttons when I used the same classes here. One sec...

ghost: image
Looks ok on my end.

peolic: Before
before
After
after

peolic:
@InfiniteTF
I thought the margin at the end was a little weird? Only saw your comment after I pushed

frontend/src/components/editCard/VoteBar.tsx
ghost: Yup, that's better. 👍

ghost (approved):

PR #140, opened 2021-10-12
Add customizable URL targets

ghost:
Replaces URL types with a `Site` entity which represents either an external website like Twitter, Wikipedia, etc. or a general link target like a Homepage. Sites can have URLs if they're concrete websites, in which case favicons are fetched. They can also have a regular expression attached which allows for cleaning pasted urls. Sites are managed the same as tag categories, admin only.

The migration is a bit sloppy since it only handles two specific types of existing urls, but those are the only ones currently used anywhere I'm aware of, so it should be fine.

PR #218, opened 2021-12-08
Refactor edit diffing

peolic:
Splitting out the edit diffing refactor + a few other changes from #140.
Probably needs a little bit more testing because I did make some changes here and there, overall is okay.

Not in the commit messages:
* Added "OTHER" to `utils.enum.ethnicityEnum` (was missing)
* Changed [`func (m *StudioEditProcessor) diffImages`](https://github.com/InfiniteTF/stash-box/blob/e0f391b41299a80a6bc04c21d83df8d312a04b51/pkg/manager/edit/studio.go#L99)'s return value to `nil`.

#140 rebased onto this branch: https://github.com/peolic/stash-box/tree/url-sites-rebased
The [`InfiniteTF/url-sites <--> peolic/url-sites-rebased` diff](https://github.com/InfiniteTF/stash-box/compare/url-sites..peolic:url-sites-rebased) is minimal so went over it, should be good.

Summary:
* Update libs
* Use `URLFragment` and `ImageFragment`
* Remove bootstrap-4 classes
* import order in api.integration_test
* Refactor edit diffing
  - StudioSelect:
	- change value shape
  - StudioForm:
    - add tabs
    - add preview of changes on confirm tab
    - rename `name` field
    - add image upload warning
  - PerformerForm:
    - fix auto breast type change on gender change
    - refactor preview of changes
* Add submission errors to edit pages
  - Prevent submission of empty tag edit
  - Replace errors in edit manager

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -31,7 +31,7 @@
     "@types/react-router-dom": "^5.1.8",
     "@typescript-eslint/eslint-plugin": "^5.4.0",
     "@typescript-eslint/parser": "^5.4.0",
-    "@vitejs/plugin-react": "1.0.5",
+    "@vitejs/plugin-react": "^1.1.1",
     "apollo": "^2.33.9",
     "eslint": "^8.2.0",
     "eslint-config-airbnb-typescript": "^16.0.0",
--- a/frontend/src/Login.tsx
+++ b/frontend/src/Login.tsx
@@ -94,7 +94,7 @@ const Login: FC = () => {
               </Link>
             </div>
           </Col>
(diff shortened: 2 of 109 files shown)


peolic:
There are circular imports between `{Performer,Scene,Studio}Form.tsx` and their corresponding `diff.ts`.
It might be causing issues for me (WSL2 hanged multiple times while working on these branches), but I'm not sure
~~It should technically be fine since they are all type imports, so that's why I'm not sure.~~ One is a type import, the other is a function.

Edit:
Well, I guess the Performer one has been there a while...

PR #220, opened 2021-12-18
Add SceneEdit/PerformerEdit/StudioEdit URL resolvers

peolic:


peolic:
Ok, maybe the resolvers are not needed? (it didn't make sense anyway)
There was an error on the staging instance, now it's gone and I can't reproduce it locally 😅

PR #222, opened 2021-12-18
Correct logic for displaying "Current Vote" info

peolic:
What I meant to do was show it to the edit submitter even when they don't have the EDIT role yet.
This corrects that.
Sorry 😅


ghost:
Duh, I missed that completely. Thanks!

PR #228, opened 2021-12-29
Fix VoteBar visibility

peolic:
* Edit status != `pending` = Nothing
* Am the user who submitted the edit (with **OR** without VOTE role) = Current Vote only
* All other users **without** `VOTE` role = Nothing
* All other users **with** `VOTE` role = Full Vote Bar

In that order


peolic:
To simplify this further,
since all users can see the Vote Tally (yes count and no count, without the names of voters),
there isn't much of a reason to hide "current vote" from non-voting users...

peolic:
Change looks good.

PR #229, opened 2021-12-30
[Hotfix] admin edit cancellation

peolic:
(verified)
PR is against `master`


peolic:
And after merging, you can use `git merge master --strategy ours` on `develop`, since this is already fixed on `develop` and we just want the two branches to merge while keeping the new version from `develop`.

PR #238, opened 2022-01-08
Assorted UI changes

peolic:
- Replace `React` imports with `FC` (and one `createContext`). Doesn't seem to affect the bundle size.
- Fix a join table name on an unused `performerQueryBuilder` function.
- Replace hardcoded routes with variables from `src/constants/route`.
- Add an `EditCard` class to edit cards.
- CI: Fail builds if `make generate` needs to be run and committed.
- Replace legacy markdown plugin [`remark-external-links`](https://github.com/remarkjs/remark-external-links) with `rehype-external-links`.


peolic:
I noticed the various edit headers (`EditHeader.tsx`, `DestroyEdit.tsx`, and `MergeEdit.tsx`) all have different styles.
For example:
	- `EditHeader.tsx` does not add a performer's disambiguation like `MergeEdit.tsx` does.
	- Each component uses a different set of elements to display similar content (`h6` vs `Row` for example)

I wonder if they could be consolidated in a way that still makes sense, unless there's a reason to keep them separate.

peolic:
At first glance it looks like some needed spacing was "lost" on merge edits between the rows of sources and target.

Looks like `h6` is setting `line-height`.

**Before** | **After**
-- | --
b | a

----

Destroy header's background was also stretched out to the entire width.

frontend/src/components/editCard/EditHeader.tsx
```diff
@@ -45,36 +45,40 @@ const EditHeader: FC<EditHeaderProps> = ({ edit }) => {
       case OperationEnum.MODIFY:
         return (
           <>
-            <Col xs={2} className="text-end">
-              Modifying {edit.target_type.toLowerCase()}:
+            <Col xs={2} className="fw-bold text-end">
+              Modifying {edit.target_type.toLowerCase()}
             </Col>
-            <Col>{renderTargetLink(edit.target)}</Col>
+            <Col className="pl-1">{renderTargetLink(edit.target)}</Col>
```
peolic: `pl-1` is bootstrap 4, but it **doesn't look like** `ps-1` is needed here if your intention is to line it up with the fields below it.
```suggestion
            <Col>{renderTargetLink(edit.target)}</Col>
```

ghost: Thanks, I was trying to line things up, but I (tried to) abandoned it.

PR #240, opened 2022-01-12
Add scene/performer draft submissions

ghost:
Drafts allow users to submit scenes/performers directly to a stash-box instance. Unlike regular scene creation this supports fingerprint submission in create edits. Images are also uploaded and attached, so scene submission should be relatively quick. If a draft isn't converted into an edit within a defined time-limit, it is automatically deleted.

Performer submission is a bit messy due to tricky field parsing, and ultimately maybe not very useful, but it does mostly work so meh.

TODO:
- [x] Remove images along with the draft if it's not used for anything else.
- [x] Add delete button to draft list.
- [x] Improve draft and fingerprint changelog rendering.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -83,7 +83,6 @@
     "rehype-external-links": "^1.0.1",
     "remark-breaks": "^3.0.1",
     "remark-gfm": "^3.0.1",
-    "remark-rehype": "10.0.0",
     "yup": "0.32.11"
   },
   "browserslist": {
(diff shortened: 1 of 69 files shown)


frontend/package.json
```diff
@@ -82,6 +82,7 @@
     "remark-breaks": "^3.0.1",
     "remark-external-links": "^9.0.1",
     "remark-gfm": "^3.0.1",
+    "remark-rehype": "10.0.0",
```
peolic: unrelated 😄

ghost: Whoops, removed, thanks!

PR #241, opened 2022-01-14
Improve performer search

### LINKED ISSUE #124: [Feature] Consider performer aliases when searching

peolic:
(2nd attempt at this)
include aliases and disambiguation (weighted)
Resolves #124 

- Hopefully this is a good query, but feel free to make changes, though I feel its a bit fragile...
but it works... I mean, it seems to work well enough for now.
- The weights are kinda random
- Some results can ~theoretically~ have a score over 1 (for example, `name similarity 1 + any alias similarity * 0.7`)
  I don't give 100% weight to the name column so exact matches are in the top results.

### DIFF
--- a/pkg/sqlx/querybuilder_performer.go
+++ b/pkg/sqlx/querybuilder_performer.go
@@ -532,12 +532,25 @@ func (qb *performerQueryBuilder) GetAllPiercings(ids []uuid.UUID) ([][]*models.B
 
 func (qb *performerQueryBuilder) SearchPerformers(term string, limit int) (models.Performers, error) {
 	query := `
-        SELECT * FROM performers
-        WHERE name % $1
-        AND similarity(name, $1) > 0.5
-        AND deleted = FALSE
-        ORDER BY similarity(name, $1) DESC
-        LIMIT $2`
+	SELECT P.* FROM performers P
+	JOIN (
+		SELECT DISTINCT ON (P.id)
+			P.id,
+			(similarity(P.name, $1) +
+			 similarity(COALESCE(PA.alias, ''), $1) * 0.7 +
+			 similarity(COALESCE(P.disambiguation, ''), $1) * 0.3
+			) AS score
+		FROM performers P
+		LEFT JOIN performer_aliases PA on PA.performer_id = P.id
+		WHERE
+			P.deleted = FALSE AND (
+				(P.name % $1 AND similarity(P.name, $1) > 0.5)
+				OR (PA.alias % $1 AND similarity(COALESCE(PA.alias, ''), $1) > 0.6)
+				OR (P.disambiguation % $1 AND similarity(COALESCE(P.disambiguation, ''), $1) > 0.7)
+			)
+	) T ON T.id = P.id
+	ORDER BY T.score DESC
+	LIMIT $2`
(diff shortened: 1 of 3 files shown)


pkg/sqlx/querybuilder_performer.go
```diff
@@ -532,25 +532,29 @@ func (qb *performerQueryBuilder) GetAllPiercings(ids []uuid.UUID) ([][]*models.B
...
-			)
-	) T ON T.id = P.id
-	ORDER BY T.score DESC
-	LIMIT $2`
+		SELECT P.* FROM performers P
+		WHERE P.id IN (
+			SELECT id FROM (
+				SELECT P.id, similarity(P.name, $1) AS similarity
+				FROM performers P
+				LEFT JOIN performer_aliases PA on PA.performer_id = P.id
```
peolic: Does it need the PA join?

pkg/sqlx/querybuilder_performer.go
```diff
@@ -532,25 +532,29 @@ func (qb *performerQueryBuilder) GetAllPiercings(ids []uuid.UUID) ([][]*models.B
...
+				WHERE P.deleted = FALSE AND (P.name % $1 AND similarity(P.name, $1) > 0.5)
+			UNION
+				SELECT P.id, (similarity(COALESCE(PA.alias, ''), $1) * 0.7) AS similarity
+				FROM performers P
+				LEFT JOIN performer_aliases PA on PA.performer_id = P.id
+				WHERE P.deleted = FALSE AND (PA.alias % $1 AND similarity(COALESCE(PA.alias, ''), $1) > 0.6)
+			UNION
+				SELECT P.id, (similarity(COALESCE(P.disambiguation, ''), $1) * 0.3) AS similarity
+				FROM performers P
+				LEFT JOIN performer_aliases PA on PA.performer_id = P.id
```
peolic: Same as above

pkg/sqlx/querybuilder_performer.go
```diff
@@ -532,25 +532,29 @@ func (qb *performerQueryBuilder) GetAllPiercings(ids []uuid.UUID) ([][]*models.B
...
+				WHERE P.deleted = FALSE AND (P.disambiguation % $1 AND similarity(COALESCE(P.disambiguation), $1) > 0.7)
+			) A
+			GROUP BY id
+			ORDER BY SUM(similarity)
```
peolic: I've hit a similar issue my first try:
You need the order to be on the final query, otherwise it doesn't keep that order.
That's why the test is failing too - it searches for some `performer-{int}`, and that should be the first match because it is an exact match.

peolic: well, correction, I'm not 100% sure it's the same issue because the query is totally different from what I tried, but it's a possibility.

PR #244, opened 2022-01-17
Fix Markdown footnotes

peolic:
Fixes footnote/back-reference links not being unique per comment, by using the comment ID (and `` `${userID}-${timestamp}` `` in the case of comment previews).
`.sr-only` class is needed because the generated footnote code uses it and Bootstrap 5 renamed the class to `.visually-hidden`.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -75,7 +75,7 @@
     "react-dom": "^17.0.2",
     "react-helmet": "^6.1.0",
     "react-hook-form": "7.19.5",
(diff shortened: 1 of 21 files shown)


ghost:
@peolic Is this ready or still draft?

peolic:
I just needed to do a final pass to see that it actually works after the release of `react-markdown`.
It is now.

ghost:
Cool, great work getting this fixed btw!

Issue #226, opened 2021-12-25
[Feature] Amending edits


peolic:
Implement the option to amend/update already-submitted edits, prior to their application/rejection.
This will help immensely in resolving conflicts (when they are identified ahead of time), correcting mistakes (typos, missing/incorrect data, switching merge direction, updating merge sources) and generally updating data due to a comment from a fellow user.
Currently the only option is to cancel and recreate the edit.

A preliminary design idea - use the entity forms (`PerformerForm`, `SceneForm`, `TagForm`, `StudioForm`).
This is open to suggestions. 

- [x] By default, the form will be populated with the edit details, merged with the entity's current values - current value, unless edit details is not null for that field (edit details override entity's current values).
- [ ] Have the option to reset specific fields to the entity's current values.
- [x] "Confirm" tab should diff the new values, either against the entity's current values, or against the edit's current details (this might be too confusing).
- [x] ~~Will require some changes to the components in order to allow passing all the "old" values (edit's `old_details`).~~
	Perhaps this will not be needed - if the "Confirm" tab will show the diff against the entity's current values, the user can determine what they need to change from that diff view.
- [ ] Merge edits are an edge case that need to be handled.
	Merge sources should be editable (removal), and the merge target should be changeable.
	I think this will add another layer of complexity to this feature, and may need to be delegated.
- [x] Create edits should be fairly straightforward, since the created entity can be constructed from `EditDetails` and passed on to the form component.
- [ ] The buttons for amending and cancelling edits should be visible (in some fashion) on the edits list.
	Currently, the cancel button is only visible on a single edit's view page, and only at the bottom of the page, which is buried in my opinion, and users easily miss it/don't even know about it.

Added by InfiniteStash:
- [x] Amending edits should remove all votes.
- [x] Limit to one amend per edit, to prevent abuse.

#365

ghost:
Sounds reasonable. I think we should be able to get away with just adding an init state to the forms, except as you say for the merge edits which needs a step to handle merge sources.

Additionally I think amending edits should remove all votes. I also think you should be limited to amending an edit once, to prevent abuse.

PR #247, opened 2022-01-19
Add initial values to forms

peolic:
Steps towards #226


peolic:
I believe this is done now, no rush to merge or anything. If this is too big, I can split out some changes I picked up along the way.
All of the forms should work as before (and seem to be).
Forms now use `defaultValues` to set default values on all the form inputs, `Controller`s used when applicable.

I didn't test the actual functionality of providing `initial` to any forms **except for `PerformerForm`** (aliases and images) because it is essential for merging.
What I mean is, it's possible more changes would have to be made for #226.

`StudioSelect` - I opted to extract the controller from the component.
I could have kept it in and provided the field name, but this way the types are more useful.
I think the recommended way to get the form control/values/state/etc, is by using `FormProvider` and `useFormContext()`.

Some of the less-related changes include:
- `boobJob` -> `breastType`
- fix types for aliases (yup/RHF now trim the aliases and transforms them to `string[]`)
- `diff.ts`/`schema.ts` setup for TagForm, like the other forms.

PR #242, opened 2022-01-15
Favorite studios and performers

ghost:
Studios and performers can be favorited by clicking the star on the entity page. Didn't have time to implement it, but the idea is that you can filter entity lists and edits, by fav status.

TODO:
- [x] Filter performer list by fav status
- [x] Filter studio list by fav status
- [x] Filter edits by performer favoriting
- [x] Filter edits by studio favoriting

PR #253, opened 2022-01-20
UI Fixes

peolic:
If this PR should be against master, we can switch it.

- Fix performer fuzzy date compare
- Fix destructive edit expiration:
	only show the minimum voting period when an edit is destructive and has passed the threshold for immediate edit application, but has not been open for the minimum voting period.


peolic:
https://github.com/stashapp/stash-box/pull/242#issuecomment-1015253403
this turned out to be wrong, margin-top: auto just glued it to the bottom, so on some performer pages it doesn't line up.

PR #271, opened 2022-01-27
Performer form/merge updates

peolic:
- _Update performer rename help message:_
	This may be the most confusing part for newcomers when editing performers.
	I believe it was suggested to provide an example, but I didn't have an idea for how to do that,
	so I just updated the text and tried to explain it more plainly.
- _Add help string to performer merging aliases:_
	Less of an issue, but probably best to have some explanation next to that checkbox as well.
	Same as above about providing an example.
- _Set merge checkbox to false if all source names match:_
	When merging `A` with `A` (or `A `), there's no reason for it to be enabled by default. That's all.
	Could also be expanded to be a case-**in*sensitive compare.
- _Check update aliases by default on performer rename:_
	Like the first commit above, this has been confusing.
	It seems most edits do need it enabled, you only need to uncheck this on rare occasions (as said, typos).
	Implementation detail - In order for the form callback to not send this value as true when the name isn't changed I opted to use `useEffect` instead of changing the default state to true. It works, but if you have a better idea, go for it.
- _Consider performer renames as destructive:_
	I'm not sure about this commit, but I included it anyway.

### DIFF
--- a/frontend/src/pages/performers/PerformerMerge.tsx
+++ b/frontend/src/pages/performers/PerformerMerge.tsx
@@ -38,15 +38,14 @@ const PerformerMerge: FC<Props> = ({ performer }) => {
     },
   });
 
-  useEffect(() => {
-    if (!mergeActive) return;
+  const toggleMerge = () => {
+    setMergeActive(true);
     const sameName = mergeSources.every(
       ({ name }) => name.trim() === performer.name.trim()
     );
     // Don't update aliases by default if the names match
     setAliasUpdating(!sameName);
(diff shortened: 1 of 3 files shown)


frontend/src/pages/performers/PerformerMerge.tsx
```diff
@@ -38,15 +38,14 @@ const PerformerMerge: FC<Props> = ({ performer }) => {
     },
   });
 
-  useEffect(() => {
-    if (!mergeActive) return;
+  const toggleMerge = () => {
+    setMergeActive(true);
```
peolic, line 42: right! this makes more sense 😅

PR #287, opened 2022-02-03
Improve error warnings in forms

ghost:


frontend/src/pages/scenes/sceneForm/SceneForm.tsx
```diff
@@ -229,12 +229,16 @@ const SceneForm: FC<SceneProps> = ({ scene, initial, callback, saving }) => {
...
-    errors.title?.message,
-    errors.date?.message,
-    errors.duration?.message,
-    errors.studio?.id?.message,
-  ]);
+  const metadataErrors = [
+    { error: errors.title?.message, tab: "details" },
+    { error: errors.date?.message, tab: "details" },
+    { error: errors.duration?.message, tab: "details" },
+    { error: errors.studio?.id?.message, tab: "details" },
```
peolic, line 236: I found this does not actually work, [but this does](https://github.com/stashapp/stash-box/pull/247/commits/20b23e5ad86658cac375e33b949dcd0a9111fc93).
There's definitely some weirdness with the type generated by `yup`, maybe v1.0 (still beta) will fix that, as it overhauls the types.

ghost: I think this might only apply when using a controller? In any case, the current approach works fine, it's what is in use in production. I'll take a look at it when merging your PR though.

PR #308, opened 2022-02-12
Hide "Favorites" edit filter when a fixed type is provided

### LINKED ISSUE #302: [Feature] Favorites toggle serves no function on the studio 'Edits' tab
**Is your feature request related to a problem?**
The 'Favorites' switch serves no function on the studio 'Edits' tab and could be removed to clean up the UI.

It could be removed.

Schermafdruk_2022-02-11_18-37-26

peolic:
If a fixed type filter is provided, there's no use for filtering favorites (currently?)
Resolves #302

### DIFF
--- a/frontend/src/components/list/EditList.tsx
+++ b/frontend/src/components/list/EditList.tsx
@@ -49,6 +49,7 @@ const EditListComponent: FC<EditsProps> = ({
     type,
     status,
     operation,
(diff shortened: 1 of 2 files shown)


ghost:
Changed it to hide the fav button if an `id` is set. End result is the same for now, but might be relevant in the future.

PR #314, opened 2022-02-14
Partial scene date

peolic:
~~As stated on Discord, still need to verify drafts work.~~
Draft->Edit works as intended.


frontend/src/pages/scenes/sceneForm/SceneForm.tsx
```diff
@@ -273,6 +278,10 @@ const SceneForm: FC<SceneProps> = ({ scene, initial, callback, saving }) => {
               <Form.Control.Feedback type="invalid">
                 {errors?.date?.message}
               </Form.Control.Feedback>
+              {/* <Form.Text>
+                If the precise date is unknown the day and/or month can be
+                omitted.
+              </Form.Text> */}
```
peolic, line 297: Copied this from `PerformerForm`, but there isn't enough space for this text.
The error message for invalid date is kind of too long now, as well.

pkg/models/model_scene.go
```diff
@@ -214,8 +215,26 @@ func CreateScenePerformers(sceneID uuid.UUID, appearances []*PerformerAppearance
...
-	p.Date = SQLiteDate{String: date, Valid: true}
+func (p *Scene) setDate(fuzzyDate FuzzyDateInput) {
+	p.Date = SQLiteDate{String: fuzzyDate.Date, Valid: fuzzyDate.Date != ""}
+	p.DateAccuracy = sql.NullString{String: fuzzyDate.Accuracy.String(), Valid: fuzzyDate.Date != ""}
+}
+
+func (p Scene) ResolveDate() *FuzzyDate {
+	if !p.Date.Valid {
+		return nil
+	}
```
peolic, line 226: This change might affect Stash, if Stash expects date to never be null.
However, the schema for Stash-Box already has the date field defined as nullable.

peolic:
Stash-Box end works.
This, however, will require an update to Stash because date is no longer a String.
https://github.com/stashapp/stash/blob/c62b0b28ed27d8d37b252dd684998db7c6e28ee1/graphql/stash-box/query.graphql#L94-L99
```gql
date
```
vs
```gql
date {
  date
  accuracy
  # or
  ...FuzzyDateFragment
}
```

I don't particularly like this, but I could keep `date` a String and add a `date_accuracy`, which _is_ backwards compatible.
I would rather have `date` be similar to the performer birthdate schema, I would have to see if it's possible to have a union of a scalar (String) with an interface (FuzzyDate) and return a different type based on the request.

peolic:
I added `DateFuzzy` and deprecated `Date` instead. This is not very elegant, but it's backwards compatible.

ghost (approved):

PR #323, opened 2022-02-16
Match scene draft URL site using regex

peolic:
Not all URLs are studio URLs, basically. Some users have indexer URLs which can be matched.
~~Draft because I didn't yet have time to test this.~~


peolic:
Just realized it should check that the site's valid types contain `SCENE` as well.

PR #363, opened 2022-04-26
Don't add identical performer name to aliases on merge

peolic:
Merging `Adam` into `Adam` should not add alias `Adam`.


peolic:
@InfiniteTF I think the merge removed the main part of this PR (the `.filter()` on the aliases array)

PR #395, opened 2022-05-21
StudioSelect parent studio

peolic:
found this branch I had forgotten about, but was practically ready.

display parent studio in Studio Select component:

image

existing (scene) studio is supported:

image

### DIFF
--- a/frontend/src/App.scss
+++ b/frontend/src/App.scss
@@ -17,6 +17,7 @@
 @import "./components/performerSelect/styles";
 @import "./components/sceneCard/styles";
 @import "./components/searchField/styles";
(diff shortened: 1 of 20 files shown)


peolic:
alternatively this can be `Studio Name • Parent Name` ([bullet](https://unicode-table.com/en/2022/))
not sure which is more clear.

ghost:
> alternatively this can be `Studio Name • Parent Name` ([bullet](https://unicode-table.com/en/2022/)) not sure which is more clear.

I switched it to bullets so it's consistent with the studios page list. Not entirely happy with it, but we can revisit that later.

PR #398, opened 2022-05-21
Add useQueryParams hook

ghost:
Replaces all query param handling with a hook. Fixes weirdness in the studio filter on the performer page.


peolic:
Interacted with it a little bit, works nicely!
Love the added feature to the performer studio filter (long URLs here we come 😄).

PR #420, opened 2022-05-28
Performer form fixes

### LINKED ISSUE #419: [Bug] Performer merge alias suggestion broken
**Describe the bug**
The alias input on PerformerForm does not suggest aliases anymore.

**To Reproduce**
1. Merge performer Bella into performer Abby
2. Remove alias "Bella".
3. Click the alias field to get suggestions.
4. "Bella" does not show up.

**Additional context**
Most likely due to `<MultiSelect>` component's `value` prop being updated by the parent component when `onChange` is triggered.
Maybe re-renders? Not quite sure as to the solution at the time of writing.

peolic:
1. Add disambiguation on performer merge sources selection.
2. Make performer gender required again.
	(`null` = "Unknown" option was pre-selected instead of the "Select gender..." option)
3. Fix #419 for performer and tag forms.
	[I seem to have misunderstood how `MultiSelect` works](https://github.com/stashapp/stash-box/pull/247/commits/011be7a85af360911fb9a83ba207b3fa51c6e0a1#diff-880813649763b2f673efbf4f63885ffe1598d32f84016c1489a4b3d858401d61), so the (initial) values being passed to the component via a property were linked to RHF and getting updated with every change to the aliases.

### DIFF
--- a/frontend/src/components/multiSelect/MultiSelect.tsx
+++ b/frontend/src/components/multiSelect/MultiSelect.tsx
@@ -3,7 +3,7 @@ import CreatableSelect from "react-select/creatable";
 import { OnChangeValue } from "react-select";
 
 interface MultiSelectProps {
-  values: string[];
+  initialValues: string[];
   onChange: (values: string[]) => void;
   placeholder?: string;
 }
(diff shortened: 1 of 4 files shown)


peolic:
✔️ ~~Would like to fix #419 (Performer merge alias suggestion broken) too, if I/we can.~~

PR #426, opened 2022-06-03
Refactor backend "/*" routes

peolic:
This initially began as a PR to fix usernames with `.` in them causing 404 errors on refresh (navigation without React), so the first commit alone fixes that.
`/static` was not being used. I assume it was the path of static assets with the original react/webpack config.
Since `vite` has been in use, `/assets` is the static assets folder.

The second commit refactors route handling a little bit, putting `/login`, `/logout`, `/image`, `/assets` and `/*` (the rest of the app/fallback) under one encompassing router.
This also makes it easier to reuse the `/assets` route handler if other routes are needed in the future.
I also added `/favicon.ico` and `/manifest.json`, as those files are requested by the browser.

The variable/function names and file name may need to be tweaked to better represent its routes.


pkg/api/routes_root.go
```diff
@@ -0,0 +1,66 @@
...
+	index []byte
+}
+
+func (rr rootRoutes) Routes() chi.Router {
+	rr.index = getIndex(rr.ui)
+
+	r := chi.NewRouter()
+
+	// session handlers
+	r.Post("/login", handleLogin)
```
peolic, line 25: Note: This is a change from the original code, which used `r.HandleFunc` and returned the index page if `r.Method == http.MethodGet`.
Since login is a POST form, makes more sense to use `r.Post`, and let the other cases fallback to the `/*` handler.

peolic:
Verified that the affected routes work as expected.
Reminding once again, since I'm still unsure about them:
>**The variable/function names and file name may need to be tweaked to better represent its routes.**

ghost:
Looks great!

PR #443, opened 2022-07-10
Frontend README update to add PR checks list

JackDawson94:
Got frustrated that there is no PR doc for people who only contribute to frontend.
So I added a few lines to the README

### DIFF
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -16,14 +16,14 @@ yarn
 ```
 
 ## GraphQL development
-If any queries/mutations or the schema on the server is updated, the Typescript types can be updated with: 
+If any queries/mutations or the schema on the server is updated, the Typescript types can be updated with:
 ```shell
 yarn generate
 ```
 
 ## Running
 
-### Local development server 
+### Local development server
 
 The API key can be set in the environment configuration. To do so, you will need to initialize the environment configuration:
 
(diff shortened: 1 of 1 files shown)


frontend/README.md
```diff
@@ -56,5 +56,29 @@ yarn format
...
+Run the code formatter:
+
+```shell
+yarn format
+```
+
+Run the typescript compiler checks:
+
+```shell
+tsc --noEmit
```
mmenanno: Should there be a note added as to where this comes from? Running this on my mac I get:
`zsh: command not found: tsc`

JackDawson94: Yeah I agree, I'm not sure how to address it. It comes with installing Typescript. Which isn't actually a dependency of the UI project.
So it's only necessary to run tsc...

peolic: > It comes with installing Typescript. Which isn't actually a dependency of the UI project.

[It is a `devDependency`](https://github.com/stashapp/stash-box/blob/bae0400549c51ec055920ea326ed736ba8f12987/frontend/package.json#L53), and the solution should be `yarn tsc --noEmit` or creating a separate entry in `scripts` for it.

JackDawson94:
Updated based on your comments.

All the commands are packaged as "yarn validate" already. So I just added this to the Readme.

PR #448, opened 2022-07-20
Join multiple capture groups on URL input

mmenanno:
This idea was came up for me when looking at the URLs for `theNude`. You will often see a URL like this `https://www.thenude.com/Kelly%20Nichols_48976.htm`. However everything between `.com/` and `_` is garbage and doesn't matter to the site. You can change that URL to `https://www.thenude.com/sdkhfkdsahdfkjwhedsfkj_48976.htm` or `https://www.thenude.com/_48976.htm` and it works exactly the same. I've seen a few other DBs catch onto this and clean their URLs so I wanted to do the same, but couldn't do that when I can only use a single capture group.

I've edited the logic here to act the same way as before if there are 0 or 1 capture groups, however if there are more than one capture groups, it joins them all together. This allows for a regex statement like `(https?:\/\/[www.thenude.com](http://www.thenude.com/)\/).*((?:_\d+).htm)` to clean up links for `theNude` when inputted. 

Sample of this code in action running my PR locally:
example

### DIFF
--- a/frontend/src/components/urlInput/urlInput.tsx
+++ b/frontend/src/components/urlInput/urlInput.tsx
@@ -50,7 +50,12 @@ const URLInput: FC<URLInputProps> = ({ control, type, errors }) => {
     const regex = new RegExp(regexStr);
     const match = regex.exec(url);
 
-    return match?.[1];
+    if (match.length > 1) {
+      match.shift();
+      return match.join("");
+    } else {
+      return match?.[1];
+    }
   };
 
   const handleAdd = () => {
--- a/frontend/src/components/urlInput/urlInput.tsx
+++ b/frontend/src/components/urlInput/urlInput.tsx
@@ -50,11 +50,11 @@ const URLInput: FC<URLInputProps> = ({ control, type, errors }) => {
     const regex = new RegExp(regexStr);
     const match = regex.exec(url);
 
-    if (match.length > 1) {
+    if (match == null || match.length < 2) {
+      return match?.[1];
+    } else {
       match.shift();
       return match.join("");
-    } else {
-      return match?.[1];
     }
   };
 


frontend/src/components/urlInput/urlInput.tsx
```diff
@@ -50,7 +50,12 @@ const URLInput: FC<URLInputProps> = ({ control, type, errors }) => {
     const regex = new RegExp(regexStr);
     const match = regex.exec(url);
 
-    return match?.[1];
+    if (match.length > 1) {
+      match.shift();
```
mmenanno, line 56: The first match is just the full matched string, everything after it is the capture groups. So we drop the first item since it is not relevant to our join.

peolic:
Your example RegExp got malformed by copy-pasting into GitHub, I assume [this is what you meant](https://regex101.com/r/zUgHi2/1)?
```regexp
(https?:\/\/www.thenude.com\/).*((?:_\d+).htm)
# or
(https?://www\.thenude\.com/).*(_\d+\.htm)
```

frontend/src/components/urlInput/urlInput.tsx
peolic: there are possible improvements to this code,
but in terms of possible bug - [`match` could be `null`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#return_value), you need to use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) to prevent an exception:
```ts
    if (match?.length > 1) {
```

mmenanno:
> Your example RegExp got malformed by copy-pasting into GitHub, I assume [this is what you meant](https://regex101.com/r/zUgHi2/1)?
> 
> ```regexp
> (https?:\/\/www.thenude.com\/).*((?:_\d+).htm)
> # or
> (https?://www\.thenude\.com/).*(_\d+\.htm)
> ```

Yup the first `(https?:\/\/www.thenude.com\/).*((?:_\d+).htm)` was what I had locally

frontend/src/components/urlInput/urlInput.tsx
mmenanno: I ended up flipping the condition order and checking for null explicitly

Issue #297, opened 2022-02-09
[Bug Report] add Non-binary as an option for genders


Tweeticoats:
**Describe the bug**
Currently the list of genders is: Unknown, Female, Male, TransFemale, TransMale and intersex.
Non-binary does not exist in an option in stash-box but is an option in stash.

**Additional context**
Are there other genders we need to consider?
Might be worth asking for suggestions on what we should store

Flashy78:
This can be closed.

PR #452, opened 2022-07-22
Add gender Nonbinary

JackDawson94:
Aligning StashBox on some of Stash's gender features:
- add the "Non Binary" gender for performers
- gender is displayed as tooltip on the gender Icon

Solves https://github.com/stashapp/stash-box/issues/297


frontend/src/pages/performers/performerForm/PerformerForm.tsx
```diff
@@ -47,14 +49,14 @@ type OptionEnum = {
...
 };
 
+const genderOptions = Object.keys(GenderEnum).map((g) => ({
+  value: g,
+  label: GenderTypes[g as GenderEnum],
+}));
+
 const GENDER: OptionEnum[] = [
   { value: "", label: "Select gender...", disabled: true },
-  { value: "null", label: "Unknown" },
```
peolic, line 52: 1. I don't think `GenderEnum` contains a mapping from `null` to "Unknown" and neither does `GenderTypes`.
2. My guess is the specific order of options was intentional here, unselected and unknown at the top, then the rest (maybe) in order of most-to-least prevalent genders. (my wording is a bit weird, what I mean is, Female is probably more common than Male, etc.)

I did not review the entire PR, just this as it caught my attention.

JackDawson94: Makes sense, will revert this change. Wanted to reduce the reliance on hardcoded gender enums, but from your description, it makes sense to have it hardcoded in this location.

JackDawson94: @peolic  just checked, the order does not change anything, you can take a look at StashDB, it's rendered in alphabetical order.
Will revert the delete of "null" because that one is necessary, but keep the spread syntax

JackDawson94:
Can't fix the remaining linting errors, it's a dependency issue. Out of scope for this PR.

PR #457, opened 2022-07-29
Fix draft update metadata removal

ghost:
Should fix the issue with urls being nuked by scene updates. I've not done anything about tags since I'm not sure if merging tags is better than overwriting.


mmenanno:
> I've not done anything about tags since I'm not sure if merging tags is better than overwriting.

I think tags should be included as well, if the user doesn't have some of the already included tags in their local instance, they will be removed when they shouldn't be. Easier to remove any unwanted tags manually since it just requires clicking an X than adding ones back that shouldn't have been removed since it involves typing

PR #481, opened 2022-08-23
[WIP] My scenes / fingerprints

JackDawson94:
### Description
New interface allowing users to manage the scenes they have submitted fingerprints for.
There they will be able to see a list of the scenes they contributed to, and if they believe one is incorrect, remove the incorrect fingerprint.

image

### Status (Updated 23/08)
UI is now working, with a new GQL endpoint which allows a user to gather all the fingerprints they have submitted (limited to phash for now)
Clicking on "X" will remove the PHASH fingerprint for that scene.

Next steps:
- Add support for OSHASH
- Improve UI to allow delete of multiple hashes at once


ghost:
After thinking about it for a bit I think a better approach than doing separate `MyFingerprints` and `Scene` queries is to add a `hasFingerprintSubmission` filter to `queryScenes`. This filter would join on scene_fingerprints so that you only get scenes you have submissions for. Additionally you can add a `owned` filter to the scene fingerprint resolver, so it only returns your own fingerprint submissions. This would get you a paginated list of scenes, with the fingerprints you care about, in a single query. Updating the cache based on deletes should be pretty simple from there.

JackDawson94:
Scrapping this PR, will submit a new one and do this in stages.

PR #495, opened 2022-09-18
Add missing phash warning to scene submission

Flashy78:
When a user submits a scene without a phash, they are given a warning at the top ~~along with a link to the guidelines explaining phash~~.
The current scene submission page is missing the hashes, so I added them and added in red MISSING text when the phash isn't there.
These are just warnings and do not stop you from submitting the scene.

Never used React before, so please let me know how I could refactor this to have better structure.

image


ghost:
Thanks for the PR. I stripped out the fingerprint addition to the form schema since that has potentially bad side-effects for scene updates. The header alert should hopefully be enough anyway.

Flashy78:
Thanks for going through it!

Issue #300, opened 2022-02-11
[Feature] Add the ability to show 'Pending' edits that haven't received your vote yet


kronkels:
**Is your feature request related to a problem?**
Currently it is not possible to show the pending edits you haven't voted on yet.

**Describe the solution you'd like**
Add a filter to display only the pending edits you have not voted on yet.

Flashy78:
This can be closed.

PR #504, opened 2022-09-22
Added user voted filter to edit list

Flashy78:
voted

Added a filter that shows up when viewing the edit queue, or another user's edits. It's hidden when viewing your own edits.
The default has been set as Not Yet Voted, while the currently functionality can be had by selecting All.

This will make going through the queue easier, and allow folks to vote No more often, since they can now easily view their No votes and not worry that they've clogged up an edit for a small reason.

I don't have a large enough dataset to really test if the Not Yet Voted WHERE NOT EXISTS clause causes any major slowdowns, maybe it's best to add indices to edit_votes for edit_id/user_id and user_id/vote? Although that may cause the vote update query to run slower.


ghost:
Resolves #300

PR #523, opened 2022-10-12
fix example regex in form text

aghoulcoder:
**EDIT:** *⚠️ This is a WIP. There are various issues with how regexps are handled and presented to the user. I'm trying to untangle it a bit.*
Periods must be escaped to be interpreted literally.

P.S. I just started using stash, so here's a trivial fix to say "hello"👋.

### DIFF
--- a/frontend/src/pages/sites/siteForm/SiteForm.tsx
+++ b/frontend/src/pages/sites/siteForm/SiteForm.tsx
@@ -98,7 +98,7 @@ const SiteForm: FC<SiteProps> = ({ site, callback }) => {
         <Form.Text>
           A regular expression that will be optionally used to clean links. Must
           contain a capture group of the portion of the URL that is considered
-          valid. For instance: <code>(https://example.org/.*)\??</code> which
+          valid. For instance: <code>(https://example\.org/.*)\??</code> which
           will capture everything before the first question mark.
         </Form.Text>
       </Form.Group>


aghoulcoder:
Actually this is still bad because `.*` will match everything, making the subsequent `\??` pointless.
This is more meaningful `(https://example.org/.*)\?.*` but I'm not sure exactly what the *intention* of the original example is.

I know this may seem nitpicky, but this is a help text and it ought to be correct, otherwise it is more misleading than helpful for a user who's not experienced with regex. So it is worth clarifying.

aghoulcoder:
This whole help text is very *unhelpful*. It's not clear what the regex does. It says it's used to "clean links" but that in itself is unclear. From interacting with the forms it's clear that it's supposed to block the usage of certain urls and also automatically detect the type of "site" based on an input url. These behaviors are not described by the phrase "clean links". Also, the enforcement of a regex for a particular site can be bypassed by selecting any other site from the drop-down menu.

Basically, this whole regex for sites functionality is a mess. If and when I understand it better I'll try to fix it.

I guess this patch can be merged regardless.

aghoulcoder:
**NOTE:** The regular expressions are used to clean the url and match a site. This all happens in the frontend using javascript's `RegExp` object and its functions. In JavaScript, forward slashes must be *escaped* [according to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#escaping):

> Similarly, if you're writing a regular expression literal and need to match a slash ("/"), you need to escape that (otherwise, it terminates the pattern). For instance, to search for the string "/example/" followed by one or more alphabetic characters, you'd use /\/example\/[a-z]+/i—the backslashes before each slash make them literal.

But the reason nobody noticed this is that apparently, chrome, firefox, and nodejs are all permissive about this. For example:

```js
RegExp("/boobs/").test("example.com/boobs/")
true
RegExp("\/boobs\/").test("example.com/boobs/")
true
RegExp("\/boobs\/").test("example.com/not-boobs/")
false
```

Still, [regex101.com](https://regex101.com/) is strict about this when selecting ECMAScript(JavaScript). The result is that people I've talked to have been using Golang in regex101 to test their regexps! So much confusion... This all needs to be sorted out :)

Also dropping these here for my own reference: 
https://urlregex.com/
https://regex101.com/r/NpFACM/1

aghoulcoder:
I made some more changes and will open a new pull request to avoid confusion.

PR #537, opened 2022-10-18
simple Makefile cleanup

aghoulcoder:
I've done some initial cleanup of the `Makefile`. In short, I have:

* made the default target do something useful (build  the application).
* grouped all the `.PHONY` declarations together so we can see at a glance all the make targets, and added a missing `.PHONY` for `pre-ui`. There are still some targets that need `.PHONY` declarations but I wasn't able to test those yet.
* added a `clean` target.

I wanted to open a pull request early to get feedback (and hopefully get this merged) because I would like to clean up the `Makefile` even more before I add more things like a `make debian` for building `.deb` packages with docker. Eventually this can even be a GitHub action and you can automatically have `.deb` and `.rpm` packages in your GitHub releases.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,7 @@ generate: generate-backend generate-ui
 	ui-start \
 	ui-fmt \
 	ui-validate \
+	pre-ui \
 	clean
 
 clean:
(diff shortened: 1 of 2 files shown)


erri120:
Just a heads-up: this will conflict with #543 if both are merged

PR #543, opened 2022-10-25
Image storage overhaul

erri120:
## Overview

This PR overhauls how images are stored. It simplifies the storage backend code and removes the discrepancies found between the current file and S3 backend.

## Changes

This PR adds new options for storing images:

- [x] `images.format`: allows specifying what format images should be stored in (independent of resizing)
  - [x] `png`: uses `image/png`.
  - [x] `jpeg` uses `image/jpeg`.
  - [x] `webp` uses [`holesa-team/go-webp`](https://github.com/kolesa-team/go-webp) (default).
  - [x] ~~`avif` uses [`Kagami/go-avif`](https://github.com/Kagami/go-avif).~~ (scraped for now)
- [x] `images.max_width` and `images.max_height`:
  - The S3 section had `s3.max_dimension` which will resize the image if the biggest side (width or height) exceeds it. This is not available for file storage, so I wanted to fix that.
  - These settings also work differently depending on the orientation of the image:
    - Horizontal images are resized based on `images.max_width`, can be disabled by setting it to `0`.
    - Vertical images are resized based on `images.max_height`, can be disabled by setting it to `0`.
- [x] `images.filter`: allows specifying the resampling filter used for resizing. This doesn't expose all filters available in [disintegration/imaging](https://github.com/disintegration/imaging#image-resizing), only the following:
  - `lanczos`: slowest but highest quality.
  - `mitchell_netravali`: (default, was previous default).
  - `linear`
  - `box`
  - `nearest_neighbor`: fastest but slowest quality.

## BREAKING CHANGES

- `s3.max_dimension`: will be removed in favor of `images.max_width` and `images.max_height`.
- new image paths (see [below](#new-image-paths))

### New Image Paths

Previously, the file backend stores images based on their MD5 checksum in the configured image folder: `{imageDir}/{checksum}` and serves images as files at `/images/{checksum}`.

Meanwhile, the S3 backend saved images with a chopped up version of the UUID: `id[0:2] + "/" + id[2:4] + "/" + id` (example: `a8/95/a89548a5-3a0b-4ef9-bce9-1b3d5c79b2ee`). The S3 backend also had custom logic for _resized_ images. Images were resized if the image was bigger than the configured `s3.max_dimension` value. If that was the case, the S3 backend would resize the image and upload **both** original and resized image. The original image was saved with the schema explained above, but the resized image would have a new path:
https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/image/s3.go#L45-L47

This PR changed the paths so that both storage backends saved images with the same file name based on the UUID: `id[0:2] + "/" + id[2:4] + "/" + id`. Here is a list of **breaking changes** depending on the storage backend:
- S3:
  - the URL for existing S3 backends has not changed, however `-resized` images will no longer be served and should be deleted
- file:
  - existing images have to be moved to their new location
  - the URL for new images will be `/images/{uuid}`
  - new images will be saved in `{imageDir}/{id[0:2] + "/" + id[2:4] + "/" + id}`

## Reasons to like this PR

- storage backend implementations have become a lot simpler
- settings for image resizing and encoding have been make accessible to the host and are independent of the storage backend
- new image customization allows host to configure depending on their storage situation (not everyone has infinite storage)
- new image options have _sensible_ defaults
- future proofs image storage with “modern” formats (WebP ~~& AVIF~~)

## Reasons to dislike this PR

- big PR with lots of breaking changes
- adds new go modules that require build dependencies:
  - [kolesa-team/go-webp](https://github.com/kolesa-team/go-webp#requirements) requires `libwebp`
  ~~- [Kagami/go-avif](https://github.com/Kagami/go-avif#requirements) requires `libaom`~~
- no `passthrough` option when an image doesn't need to be resized, there might be generation loss due to re-encoding of lossy data

## Conclusion

Although this PR is very daunting and can break a lot of things, I believe sorting this situation out and cleaning up the image storage code is better to do now, then later.


erri120:
## Issue

While working on this, I found a major issue that needs to be solved:

Currently, the file storage backend saves images in `{imageDir}/{checksum}`:
https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/image/utils.go#L97-L99
which allows for easy file serving:
https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/api/urlbuilders/image.go#L15-L17

Meanwhile, the S3 storage backend saves images with the MD5 checksum of either `image.ID.String()` or `image.ID.String() + "-resized"`, if the image is resized:
https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/image/s3.go#L45
https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/image/s3.go#L73
https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/api/urlbuilders/s3image.go#L21-L32

This discrepancy between the two storage backends should be resolved sooner than later, however will be a **massive** breaking change.

(btw, I hope you can see the bug with the S3 URL builder. Since the builder looks at the _current_ config, if you changed the config in the past it can load the original image and not the resized one)

## Possible Solution

The fact that the S3 storage backend currently keeps the original image makes the code instantly more complex. You can see from the URL builder that it has to check if the image is resized according to the settings and then serve the right image. I intended to make this an option `images.keep_original`, however I strongly believe the original image should not be kept. With this you don't need any checks and just have one static URL for the image.

Removing the original also means that the entire function for adding new images has to be reworked:

https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/image/service.go#L18

This function should manipulate the image, discard the original image, and generate a checksum with the manipulated image. The only logic inside backends will then be to save the image, nothing more.

erri120:
I've found another weird piece of code:

https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/image/service.go#L24-L33

This “fix” was implemented because users reported an issue with their AdBlocker because some UUIDs ended up starting with `ad` and got blocked: `https://example.com/ad/xx/xxxxxxxxxxxxxxx` (https://discord.com/channels/559159668438728723/642050893549928449/831269018454851644)

Here is again the discrepancy between file storage and S3 storage:
- The file storage backend saves images in the configured path, with the file name being the checksum. This allows for an easy router implementation: https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/api/routes_image.go#L31
- The S3 storage backend however chops up the UUID of the image: https://github.com/stashapp/stash-box/blob/546ba8de9cc04c9c861ea92ac2ccd7e11072e8d9/pkg/api/urlbuilders/s3image.go#L31

The file storage backend should also use the chopped up UUID instead of the checksum as a file name. This kind of file structure is better for the file system but requires a breaking change.

erri120:
I don't understand why the build fails in the CI. It works fine on my machine and in my docker container.

PR #555, opened 2022-11-14
Measurements parse

JackDawson94:
Even if Stash doesn't use the same data format as StashBox, we should at least be able to re-parse data in the format Stash uses when it imports data from StashBox.

So, I updated the Performer Draft parser, allows it to parse performer's measurements.
Parsing is based on the syntax Stash uses when grabbing the measurements in StashBox.

Example: 32A-25-25


JackDawson94:
@InfiniteStash  After review of your refactor, all my tests are working 😊

PR #564, opened 2022-11-22
Now parses aliases

JackDawson94:
Allows the Aliases to be parsed when submitting a performer from Stash.
Only uses "," as a separator because that's the separator Stash uses when **scrapping** from Stashbox


JackDawson94:
@InfiniteStash  after review of your refactor, all my tests are still working !

PR #591, opened 2023-02-01
Compose modernization and added ref to Stash pHash compat img.

Leopere:
- removed redundant `container_name:`
- removed deprecated `link:`
- added image: stashapp/postgres:latest to postgres image.


Leopere:
Going forward with these tweaks you could run `docker compose build postgres && docker compose push postgres` so long as you have push perms on the official repo until we get some gitops setup for it.

treyt02 (approved):

PR #597, opened 2023-02-10
Allow Multiple Edits and Reset Votes

### LINKED ISSUE #593: [Feature] Allow Multiple Edits and Reset Votes
One issue that keeps frustrating users is that they can only edit their submissions once. This is most often seen with new contributors who often need guidance with their submissions before they learn the ropes. They can get discouraged because their submission was downvoted because they could only make the first suggested change to it, or they have to cancel their submission and re-do their effort.

It was mentioned in https://github.com/stashapp/stash-box/issues/226 that users should be limited to a single edit to prevent abuse. The assumption is that this was because votes are not reset when an edit is updated.

I propose changing it so that previous votes are no longer counted when an edit is updated, and the limit on the number of updates is removed.

**Updated Implementation Ideas**

Thought about it more and I think it's just best to do the filtering in the database without trying to hack it. That makes it consistent and will work for other voting queries such as the user voted filtering.

**Initial Implementation Ideas**

Modify https://github.com/stashapp/stash-box/blob/master/pkg/api/resolver_model_edit.go#L314 to filter out votes cast after the `updated` time. This makes it easy on the database queries by not introducing filtering on the large `edit_votes` table, and the amount of votes for an edit shouldn't be that large that filtering in code has any negative affect.
However if a user is sitting on the edit queue webpage and hasn't refreshed in a while, they may not get the latest changes to an edit when they submit their vote and it could be counted towards the updated version.
On the frontend it complicates the Votes section of an edit. With the above approach, it would only show votes for the current update. Maybe a second section could be added showing previous votes. 

This also complicates the user voted filter, as they could have previously voted on an edit, but the updated edit should show as if they haven't voted yet. Not sure how to handle this, it may make the approach above with filtering in code not viable.

Thoughts and suggestions?

Flashy78:
Resolves #593 
This removes the limit of updating an edit more than once.

When getting the votes for an edit, if the edit has been updated, it ignores any votes that were cast before the update.

That would have made the existing `votes` column on the edit table harder to manage with the database trigger and such. Since several places were already just tallying the votes each time, I decided to remove the trigger and that column and just always tally the current votes.

I will add comments in the code where I did things I'm not entirely sure about.


pkg/api/resolver_model_edit.go
```diff
@@ -58,9 +58,27 @@ func (r *editResolver) Expires(ctx context.Context, obj *models.Edit) (*time.Tim
 		startTime = obj.UpdatedAt.Time
 	}
 
+	fac := r.getRepoFactory(ctx)
+	qb := fac.Edit()
+	editVotes, err := qb.GetVotes(obj.ID)
```
Flashy78, line 63: This seems worrisome that I have to go back to the DB to get the votes in order to calculate the expires time for each entry. It's the exact same code as the GetVotes method further down in this file.
Is there any caching happening so that maybe it won't have to hit the database twice for the same data for every edit?

pkg/manager/cron/cron.go
```diff
@@ -41,8 +41,23 @@ func (c Cron) processEdits() {
 				voteThreshold = 1
 			}
 
-			var err error
-			if e.VoteCount >= voteThreshold {
+			editVotes, err := c.rfp.Repo().Edit().GetVotes(e.ID)
```
Flashy78, line 44: The cron now has to go and get the votes for each edit, but it's only happening every few minutes, and there shouldn't be more than reasonable amount of edits to handle each run, so I think this is ok.

pkg/manager/cron/cron.go
```diff
@@ -41,8 +41,23 @@ func (c Cron) processEdits() {
...
+			if err != nil {
+				logger.Errorf("Edit cronjob failed to fetch edit votes for edit %s: %s", e.ID.String(), err.Error())
+				return err
```
Flashy78, line 47: I'm not entirely sure that if there's an error, it will continue processing the other edits, or if it will all just crash.

pkg/sqlx/querybuilder_edit.go
```diff
@@ -410,7 +421,18 @@ func (qb *editQueryBuilder) FindCompletedEdits(votingPeriod int, minimumVotingPe
...
+					  CASE
+						WHEN vote = 'ACCEPT' THEN 1
+						WHEN vote = 'REJECT' THEN -1
+						ELSE 0
+					  END
+					) as votesum
+					FROM edit_votes
+					WHERE edit_id = edits.id
+					AND (edits.updated_at IS NULL OR edit_votes.created_at > edits.updated_at)
+				)
```
Flashy78, line 435: This appears to work as intended. The code is from the trigger function that was deleted.

Flashy78:
~~Would it make sense the tests are failing because the testing database did not get the migration?~~

Missed updating the schema version in the code.

InfiniteStash:
@Flashy78 Thanks for filing the ticket and taking a look at this.

I'm closing this in favor of an alternative PR that makes the update limit configurable, and resets the votes instead of just counting votes since the last update. While it would be nice to keep the original votes, I think having "expired" votes is more confusing than helpful.

Anyway, thanks again for getting this underway.

PR #598, opened 2023-02-10
Abstain vote

### LINKED ISSUE #588: [Feature] Ability to abstan vote as 1st vote
Currently users only have the option to `Abstain` and click `Save` after they've voted `Yes` or `No`.  With the recent addition of user vote-filtering (look ma, my queue's clean!) and the only initial offerings of Yes/No, this pressures a user toward making a binary vote when maybe they should pass.

Allow users to select `Abstan` from the onset

Personally I vote `No` then change to `Abstain`, but I'm not normal and typical people take the path of least resistance

Flashy78:
Resolves #588 

Basically removed `(userVote || vote !== VoteTypeEnum.ABSTAIN)`. Not entirely sure why userVote was here, I don't think it's needed.

### DIFF
--- a/frontend/src/components/editCard/VoteBar.tsx
+++ b/frontend/src/components/editCard/VoteBar.tsx
@@ -62,18 +62,16 @@ const VoteBar: FC<Props> = ({ edit }) => {
     <div className={CLASSNAME}>
       <div className={CLASSNAME_SAVE}>
         {currentVote}
-        {vote &&
-          vote !== userVote?.vote &&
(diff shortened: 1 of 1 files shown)


InfiniteStash:
> Basically removed `(userVote || vote !== VoteTypeEnum.ABSTAIN)`. Not entirely sure why userVote was here, I don't think it's needed.

The logic is that you can only make an abstain vote if you've already voted for something else. But yes, unnecessarily complicated I think.

PR #601, opened 2023-02-10
Add guidelines link in top menu

### LINKED ISSUE #512: [Feature] Cosmetic navbar refactor
There's a litany of guidelines for StashDB but the link is (possibly) pinned somewhere in a Discord channel. This isn't an ideal location for a project that requires contributors to be versed with the docs. There should be a clear and obvious link to these documents front-and-center for end-users.

Add a `Documentation` link to the menu bar.

**Additional context**
Adding a `Documentation` link starts to make the menu bar a bit crowded, but that can be easily solved because all-user specific menu items (Drafts, Logout, etc) should be consolidated into one group -- a user drop-down menu. See image below.

stashdb

Flashy78:
Resolves #512 

Adds an link called Guidelines that opens in a new tab, to the end of the left hand top menu items. It does not show up if the config value is empty.

Alternatively it could be added every create/edit page instead so it doesn't take up space in the top menu all the time.
Or on the edit confirm screen, if someone has less than 100 edits it could also be displayed in text asking them to ensure they've read the guidelines.


WithoutPants:
The navbar is getting a bit crowded. I think for now, this might be best suited in the Edit Confirm page.

Flashy78:
Some discussion on where folks think it should be displayed: https://discord.com/channels/559159668438728723/798641040029777980/1119309230126678076

PR #608, opened 2023-02-17
Update hair colors and fix text display on edit

### LINKED ISSUE #209: [Feature] Update language of existing hair color and add white hair
- `Brunette` and `blonde` are not gender neutral words. They're words using a feminine form (referring to women and girls distinctly).[^blond][^brown] Stashbox incorrectly uses the feminine form of said words to describe male, intersex, and unknown gender performers.
- White hair is missing

- Change `brunette` to the gender neutral word `brown`
- Change `blonde` to the gender neutral word `blond`
- Add `white` hair (Stashbox already includes grey hair -- a mixture of white and existing hair that hasn't lost pigmentation -- and white hair is the next step in the process of "graying", when all hair has lost pigmentation or is artificially made white)

**Additional context**
This isn't coming from a place of wokeness or anything like that. It's simply a proper grammar thing since Stash uses the same set of words to describe men, women, intersex, etc.

I read it might be acceptable in some English language regions to use the feminine form for males too, but it's also acceptable to use the gender neutral form in those regions. Since it is correct to use the gender neutral form in all instances, I would default to the neutral form.

[^blond]: Blond is the gender neutral term that can refer to both men or women (or intersex and unknown genders). [[Source](https://www.oxfordlearnersdictionaries.com/definition/english/blonde_1)]
[^brown]: Brunette is used for women with dark brown hair. [[Source](https://www.oxfordlearnersdictionaries.com/definition/american_english/brunette)]

Flashy78:
Resolves #209 

- Updates existing values of BRUNETTE to BROWN and BLONDE to BLOND in the DB.
- Adds White as a hair color option.
- Fixes text display of enum values on the edit diff performer pages to use the nicer display value instead of all uppercase.

This does not attempt to maintain backwards compatibility with the previous enums, so those values will no longer work in GraphQL.


InfiniteStash:
Thanks! I've backed out the enum renames to avoid causing a breaking change.

PR #635, opened 2023-04-18
include bot edits in pending count

mmenanno:
~Previously when using a bot edit, no pending edit count would show on things like the scene page. This PR fixes this by also pulling the bot edit count and combining the numbers (currently these need to be separate queries, the `queryEdits` query will either pull bot edits or regular edits but not both at once).~

~I also used `|| 0` to default the counts to 0 if they are undefined, this allowed for much simpler logic when combining them not having to deal with the potential of one of the numbers being undefined.~

~I submitted a bot edit to production and tested against that.~

This change is now done simply by allowing the absence of the is_bot field to return both true and false values.

Before my PR:
image

After my PR:
image


Flashy78:
I wonder if it would be worthwhile instead to update the `querybuilder_edit.buildQuery` to support `true, false, all` for `filter.IsBot`. That would save an extra DB call for every page view.
Since the edit queue page is already always passing true/false, we could just have `null == all`, so if it's not passed in, we don't filter and return both bot and non-bot edits.

mmenanno:
> I wonder if it would be worthwhile instead to update the `querybuilder_edit.buildQuery` to support `true, false, all` for `filter.IsBot`. That would save an extra DB call for every page view. Since the edit queue page is already always passing true/false, we could just have `null == all`, so if it's not passed in, we don't filter and return both bot and non-bot edits.

That's a good point, let me look into that

mmenanno:
I think the latest version of the change should work, but I think I need to have the backend on my PR for it to work. I don't have a lot of luck with getting the backend to build and setup locally, so if someone could test for me I'd really appreciate it.

mmenanno:
^ Confirmed on discord that the change worked 
https://discord.com/channels/559159668438728723/559169859351740418/1098074007342944356

Issue #645, opened 2023-05-11
[BUG] Create scene and create studio page broken at 8833c7b


aghoulcoder:
## Problem
Create scene and create studio page broken at 8833c7b

## To reproduce
1. Build stash-box from `master` or on commit 8833c7b
2. Run it
3. Go to http://localhost:9998/scenes/add
4. Page doesn't render. Console shows this:

```
react-dom.production.min.js:189 TypeError: The value of note could not be cast to a value that satisfies the schema type: "string". 

attempted value: undefined 

    at u8.cast (index.esm.js:617:13)
    at m8._cast (index.esm.js:1683:15)
    at m8.cast (index.esm.js:610:33)
    at SceneForm.tsx:106:33
    at Object.useMemo (react-dom.production.min.js:184:189)
    at qe.useMemo (react.production.min.js:25:208)
    at Sy (SceneForm.tsx:105:46)
    at tE (react-dom.production.min.js:167:137)
    at x4 (react-dom.production.min.js:290:337)
    at S4 (react-dom.production.min.js:280:389)
a1 @ react-dom.production.min.js:189
t.callback @ react-dom.production.min.js:189
bC @ react-dom.production.min.js:144
MC @ react-dom.production.min.js:262
v4 @ react-dom.production.min.js:260
EU @ react-dom.production.min.js:259
TU @ react-dom.production.min.js:283
Us @ react-dom.production.min.js:281
UC @ react-dom.production.min.js:273
ws @ react-dom.production.min.js:127
(anonymous) @ react-dom.production.min.js:267
index.esm.js:617 Uncaught TypeError: The value of note could not be cast to a value that satisfies the schema type: "string". 

attempted value: undefined 

    at u8.cast (index.esm.js:617:13)
    at m8._cast (index.esm.js:1683:15)
    at m8.cast (index.esm.js:610:33)
    at SceneForm.tsx:106:33
    at Object.useMemo (react-dom.production.min.js:184:189)
    at qe.useMemo (react.production.min.js:25:208)
    at Sy (SceneForm.tsx:105:46)
    at tE (react-dom.production.min.js:167:137)
    at x4 (react-dom.production.min.js:290:337)
    at S4 (react-dom.production.min.js:280:389)
cast @ index.esm.js:617
_cast @ index.esm.js:1683
cast @ index.esm.js:610
(anonymous) @ SceneForm.tsx:106
useMemo @ react-dom.production.min.js:184
qe.useMemo @ react.production.min.js:25
[...]

PR #652, opened 2023-06-03
More reliable favicon retrieval.

aghoulcoder:
Previously, the application looked for favicons in `/favicon.ico`. But favicons can be anywhere on a website and the location can be defined in HTML and JSON manifests. Websites can also have multiple favicons.

As a result, when a new site was added to stash-box which did not have a `/favicon.ico`, the site was created without a proper favicon. Additionally, since stash-box didn't check the filetype of the data returned by the `/favicon.ico` GET request, it stored various 404 and other redirected html pages.

A non-trivial amount of logic is required to discover all favicon locations, check filetypes, and sort them. It therefore makes sense to rely on a third-party package for this job. This patch uses `go.deanishe.net/favicon`.

### DIFF
--- a/pkg/image/favicon.go
+++ b/pkg/image/favicon.go
@@ -71,8 +71,10 @@ func downloadIcon(ctx context.Context, iconPath string, siteURL string) {
 		return
 	}
 
-	// Icons are sorted widest first. We currently get the first one (icons[0]).
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, icons[0].URL, nil)
+	// Icons are sorted widest first. Based on the design of the stash-box UI,
+	// it makes sense to grab the _smallest_ icon, i.e. the last one.
+	// TODO: Find the ideal size favicon for the stash-box UI and try to get the same here.
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, icons[len(icons)-1].URL, nil)
 	if err != nil {
 		return
 	}
--- a/pkg/image/favicon.go
+++ b/pkg/image/favicon.go
@@ -73,7 +73,7 @@ func downloadIcon(ctx context.Context, iconPath string, siteURL string) {
 	}
 
 	icons, err := favicon.Find(u.Scheme + "://" + u.Host)
-	if err != nil {
+	if err != nil || len(icons) < 1 {
 		return
 	}
 
--- a/pkg/image/favicon.go
+++ b/pkg/image/favicon.go
@@ -57,12 +57,6 @@ func GetSiteIcon(ctx context.Context, site models.Site) []byte {
 }
 
 func downloadIcon(ctx context.Context, iconPath string, siteURL string) {
-	out, err := os.Create(iconPath)
-	if err != nil {
-		return
-	}
-	defer out.Close()
-
 	if siteURL == "" {
 		return
 	}
@@ -87,6 +81,12 @@ func downloadIcon(ctx context.Context, iconPath string, siteURL string) {
 		return
 	}
 	defer resp.Body.Close()
+
+	out, err := os.Create(iconPath)
+	if err != nil {
+		return
+	}
+	defer out.Close()
 
 	//nolint
 	io.Copy(out, resp.Body)
--- a/pkg/image/favicon.go
+++ b/pkg/image/favicon.go
@@ -12,6 +12,7 @@ import (
 	"github.com/gofrs/uuid"
 	"github.com/stashapp/stash-box/pkg/manager/config"
 	"github.com/stashapp/stash-box/pkg/models"
+	"go.deanishe.net/favicon"
 )
 
 var iconCache = map[uuid.UUID][]byte{}
@@ -70,9 +71,14 @@ func downloadIcon(ctx context.Context, iconPath string, siteURL string) {
 	if err != nil {
 		return
 	}
-	u.Path = path.Join(u.Path, "favicon.ico")
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	icons, err := favicon.Find(u.Scheme + "://" + u.Host)
+	if err != nil {
+		return
+	}
+
+	// Icons are sorted widest first. We currently get the first one (icons[0]).
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, icons[0].URL, nil)
 	if err != nil {
 		return
 	}
--- a/pkg/image/favicon.go
+++ b/pkg/image/favicon.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"net/http"
+	"net/http/cookiejar"
 	"net/url"
 	"os"
 	"path"
@@ -13,6 +14,7 @@ import (
 	"github.com/stashapp/stash-box/pkg/manager/config"
 	"github.com/stashapp/stash-box/pkg/models"
 	"go.deanishe.net/favicon"
+	"golang.org/x/net/publicsuffix"
 )
 
 var iconCache = map[uuid.UUID][]byte{}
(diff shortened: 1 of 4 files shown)


aghoulcoder:
Merge conflict is because I wrote and tested this against `v0.4.6` instead of `master` because `master` is currently broken due to #645

pkg/image/favicon.go
```diff
@@ -70,9 +71,14 @@ func downloadIcon(ctx context.Context, iconPath string, siteURL string) {
...
-	u.Path = path.Join(u.Path, "favicon.ico")
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	icons, err := favicon.Find(u.Scheme + "://" + u.Host)
+	if err != nil {
+		return
+	}
+
+	// Icons are sorted widest first. We currently get the first one (icons[0]).
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, icons[0].URL, nil)
```
Maista6969: This will panic with a runtime error if `icons` is empty

aghoulcoder: Super! thanks.

aghoulcoder:
force-pushed after rebasing on `master` and resolving the merge conflict in `go.mod`.

WithoutPants:
I haven't been able to get this to work reliably. There's a few issues with the icon downloading code which make debugging the issues quite difficult. I think the whole thing needs a bit of improving, but probably out of scope of what you're doing.

I tried creating a new site with URL `https://twitter.com`. The icon downloading code fails silently if the `favicon_path` config field isn't populated. I then populated that field and restarted. The second issue I found was that the `downloadIcon` code creates the icon file on the filesystem regardless of whether it successfully downloads the icon. This means that if the download process fails for _any_ reason, then you have an empty file in the filesystem and you can't retrigger a download without deleting the file. I suspect that we probably shouldn't be trying to download the favicon via an http route, and there's no protection from multiple threads trying to download at once.

Finally, to the part that is the most relevant to you, the call to `favicon.Find` returns the following error `fetch page: retrieve URL: Get \"/\": stopped after 10 redirects`. This error is not logged any where and there's no indication that it failed or why without debugging. On the `master` branch, the twitter favicon is correctly downloaded and displayed.

aghoulcoder:
There's a general absence of logging throughout this codebase even though there is a `logger` package implemented and used for only a couple of things. Logging is something I would like to help improve. However, I'm not sure if I should start using the current logger package as it relies on https://github.com/sirupsen/logrus which is no longer under development.

For now I'll look into the `favicon.Find` issue.
Thanks for checking.

EDIT: There are also two important config options `logfile` and `loglevel` which are not documented.

aghoulcoder:
> the call to favicon.Find returns the following error `fetch page: retrieve URL: Get \"/\": stopped after 10 redirects`

Apparently, there's an issue with twitter.com:
```
$ curl -L https://twitter.com
curl: (47) Maximum (50) redirects followed

$ curl -v -L https://twitter.com                                                                                                                                                                                                                                  
*   Trying 104.244.42.65:443...                                                                                                                                                                                                                                                   
* Connected to twitter.com (104.244.42.65) port 443 (#0)                                                                                                                                                                                                                          
* ALPN: offers h2,http/1.1                                                                                                                                                                                                                                                        
* TLSv1.3 (OUT), TLS handshake, Client hello (1):                                                                                                                                                                                                                                 
*  CAfile: /etc/ssl/certs/ca-certificates.crt                                                                                                                                                                                                                                     
*  CApath: none                                                                                                                                                                                                                                                                   
* TLSv1.3 (IN), TLS handshake, Server hello (2):                                                                                                                                                                                                                                  
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):                                                                                                                                                                                                                          
* TLSv1.3 (IN), TLS handshake, Certificate (11):                                                                                                                                                                                                                                  
* TLSv1.3 (IN), TLS handshake, CERT verify (15):                                                                                                                                                                                                                                  
* TLSv1.3 (IN), TLS handshake, Finished (20):                                                                                                                                                                                                                                     
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):                                                                                                                                                                                                                       
* TLSv1.3 (OUT), TLS handshake, Finished (20):                                                                                                                                                                                                                                    
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384                                                                                                                                                                                                                           
* ALPN: server accepted h2                                                                                                                                                                                                                                                        
* Server certificate:                                                                                                                                                                                                                                                             
*  subject: C=US; ST=California; L=San Francisco; O=Twitter, Inc.; CN=twitter.com                                                                                                                                                                                                 
*  start date: Feb  5 00:00:00 2023 GMT                                                                                                                                                                                                                                           
*  expire date: Feb  5 23:59:59 2024 GMT                                                                                                                                                                                                                                          
*  subjectAltName: host "twitter.com" matched cert's "twitter.com"                                                                                                                                                                                                                
*  issuer: C=US; O=DigiCert Inc; CN=DigiCert TLS Hybrid ECC SHA384 2020 CA1                                                                                                                                                                                                       
*  SSL certificate verify ok.                                                                                                                                                                                                                                                     
* using HTTP/2                                                                                                                                                                                                                                                                    
* h2 [:method: GET]                                                                                                                                                                                                                                                               
* h2 [:scheme: https]                                                                                                                                                                                                                                                             
* h2 [:authority: twitter.com]                                                                                                                                                                                                                                                    
* h2 [:path: /]                                                                                                                                                                                                                                                                   
* h2 [user-agent: curl/8.1.2]                                                                                                                                                                                                                                                     
* h2 [accept: */*]                                                                                                                                                                                                                                                                
* Using Stream ID: 1 (easy handle 0x55ff0af3cf50)                                                                                                                                                                                                                                 
> GET / HTTP/2                                                                                                                                                                                                                                                                    
> Host: twitter.com                                                                                                                                                                                                                                                               
> User-Agent: curl/8.1.2                                                                                                                                                                                                                                                          
> Accept: */*                                                                                                                                                                                                                                                                     
>                                                                                                                                                                                                                                                                                 
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):                                                                                                                                                                                                                             
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):                                                                                                                                                                                                                             
* old SSL session ID is stale, removing                                                                                                                                                                                                                                           
< HTTP/2 302                                                                                                                                                                                                                                                                      
< date: Tue, 06 Jun 2023 09:18:22 GMT                                                                                                                                                                                                                                             
< perf: 7626143928                                                                                                                                                                                                                                                                
< vary: Accept                                                                                                                                                                                                                                                                    
< server: tsa_o
< location: /
< set-cookie: guest_id=v1%3A168604310292854621; Max-Age=34214400; Expires=Sat, 06 Jul 2024 09:18:22 GMT; Path=/; Domain=.twitter.com; Secure; SameSite=None
< content-type: text/plain; charset=utf-8
< x-powered-by: Express
< cache-control: no-cache, no-store, max-age=0
< content-length: 23
< x-transaction-id: 1aae4a21b4a8748b
< strict-transport-security: max-age=631138519
< x-response-time: 111
< x-connection-hash: 325f3f2bcd4b00f8806219564526f8a3783bd7130251c6000a66c8b1090e3466
< 
[continues...]
```

aghoulcoder:
Basically, the way twitter.com works is that on the first request it gives you a 302 redirect to `/` and a cookie named `guest_id`. Then you do the next request and send it the `guest_id` cookie at which point it returns 200. If you don't send it the cookie it redirects infinitely.

With curl you can store the cookie like this:
```
$ curl -c /tmp/cookies -v -L https://twitter.com
```

So that's what's happening. But I'm still not sure how to handle it.

aghoulcoder:
I think this can be solved by using the [WithClient function](https://pkg.go.dev/go.deanishe.net/favicon#WithClient) to specify an `http.Client` with a `CookieJar`. Looking into it...

aghoulcoder:
- ✅ I fixed the twitter.com redirect loop issue by using an http.client with a cookiejar. fee8a3d
- ✅ I fixed the empty file issue by moving file creation to the end of the function. f238b86
- ⏳ Logging is a problem throughout the codebase. I have opened #655 for further discussion and suggestions.
@WithoutPants

PR #705, opened 2023-08-25
Add report fingerprint functionality

WithoutPants:
Adds a `vote` field to the `FingerprintSubmission` input. This supercedes the `unmatch` field, which is now deprecated. If `vote` is positive, then the submission is treated as an ordinary submission. If `vote` is negative, then the submission is treated as a report. If `vote` is `0`, then it is treated as an unmatch operation. If `vote` is omitted altogether, then it is assumed to be a normal submission, unless `unmatch` is set to true, per existing behaviour.

Adds `reports` field to the `Fingerprint` type. This value is the number of users that have reported a fingerprint. Also adds a `user_reported` field which returns true if the current user reported the fingerprint. Scene fingerprints are now ordered by `submissions - reports`. 

The UI is changed to indicate how many reports a fingerprint has. The button to remove a user submission is shown in the appropriate place based on if the user submitted or reported the fingerprint.

image


InfiniteStash:
Looks good! Is the idea that the logic for submitting and evaluating reports stays is implemented in the stash tagging interface?

graphql/schema/types/scene.graphql
```diff
@@ -64,7 +70,9 @@ input FingerprintQueryInput {
 input FingerprintSubmission {
   scene_id: ID!
   fingerprint: FingerprintInput!
-  unmatch: Boolean
+  unmatch: Boolean @deprecated(reason: "Use `vote` with 0 instead")
+  "positive for default behaviour, negative to report as invalid, zero to remove vote"
+  vote: Int
```
InfiniteStash: This should ideally be an enum so we don't have to clamp it, but I guess that would mean duplicating the type and creating translate functions, which sounds tedious.

WithoutPants:
Had to rework this after the fingerprint normalisation.

> Looks good! Is the idea that the logic for submitting and evaluating reports stays is implemented in the stash tagging interface?

That's correct. It wouldn't be difficult to add report submission to the UI, but it feels like a potential vector of abuse to me.

I have changed the vote input to be an enum instead of an int. Should be ready for review and test.

Issue #594, opened 2023-02-08
[Feature] Confirmation on cancel


QxxxGit:
I've been tinkering with stash box and going through and creating performers etc. However, when I click on cancel in the forms there isn't a prompt so it's super easy to lose the data you've written up. I think a small confirmation prompt would help with accidental clicks.

JediMaster25:
This has happened to me a couple time when updating or closing the page by mistake. Periodically saving the data to continue where left would solve more cases.

ghost:
> One way to achieve this is by using a state management library like React's Context or Redux to store the state of the form fields that need to be checked for changes.
> 
> Here is a general outline of the steps you could follow:
> 
> 1. Create a new context or Redux store to hold the form state.
> 1. In each component that has a form with changes that need to be checked, update the state when the form values change.
> 1. Add a useEffect hook to each of these components that listens for changes to the form state, and sets a flag to indicate that the form has unsaved changes.
> 1. Modify your SubmitButton component to check this flag before allowing the user to cancel. If the flag is set, display a confirmation dialog.
> 1. Add a function to reset the form state when the user submits the form or cancels with confirmation.
> 
> Here's some sample code to illustrate the steps:
> 
> ```tsx
> // Define a new context or Redux store to hold the form state
> import { createContext, useContext } from 'react';
> 
> interface FormState {
>   // Add properties for each form field that needs to be tracked
>   name: string;
>   email: string;
>   // ...
> }
> 
> interface FormContextType {
>   formState: FormState;
>   setFormState: (state: FormState) => void;
> }
> 
> const FormContext = createContext<FormContextType>({
>   formState: {},
>   setFormState: () => {},
> });
> 
> export function useFormContext() {
>   return useContext(FormContext);
> }
> 
> // In each component with a form, update the state when the form values change
> import { useFormContext } from './form-context';
> 
> export function MyForm() {
>   const { formState, setFormState } = useFormContext();
> 
>   function handleInputChange(event: React.ChangeEvent<HTMLInputElement>) {
>     const { name, value } = event.target;
>     setFormState((prevState) => ({ ...prevState, [name]: value }));
>   }
> 
>   // Add a useEffect hook to set a flag when the form state changes
>   const [hasUnsavedChanges, setHasUnsavedChanges] = useState(false);
>   useEffect(() => {
>     setHasUnsavedChanges(true);
>   }, [formState]);
> 
>   // Add a function to reset the form state
>   function handleFormSubmit() {
[...]

PR #730, opened 2023-11-15
add native beforeUnload handlers

feederbox826:
partially addresses #594 

doesn't work with SPA navigation because

[it's "unstable"](https://www.github.com/remix-run/react-router/blob/main/packages/react-router-dom/index.tsx#L1796-L1828), reactrouter [does not intercept it](https://reactrouter.com/en/main/hooks/use-before-unload), it [does not work on SPAs](https://www.github.com/remix-run/react-router/issues/10894), the [PR](https://www.github.com/remix-run/react-router/pull/4359) ([s](https://www.github.com/remix-run/react-router/pull/4372)) are dead, is [still unstable](https://www.github.com/remix-run/react-router/issues/8139)

If someone else who is more versed and has more patience for SPAs and React wants to improve, it they are more than welcome to. In the meantime I will just [intercept navigate events](https://developer.mozilla.org/en-US/docs/Web/API/Navigation/navigate_event)

### DIFF
--- a/frontend/src/hooks/useBeforeUnload.ts
+++ b/frontend/src/hooks/useBeforeUnload.ts
@@ -1,6 +1,6 @@
 import { useEffect } from "react";
 
-export const NativeBeforeUnload = () => {
+export const useBeforeUnload = () => {
   const unloadListener = (event: BeforeUnloadEvent) => {
     event.preventDefault();
     event.returnValue = true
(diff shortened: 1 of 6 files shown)


feederbox826:
will rebase

InfiniteStash:
Looks like react-router has official support now, but it's still more tedious to use `useBlocker` than the native implementation since you have to create your own modal.

Anyway, thanks!

PR #751, opened 2024-02-03
Add fingerprint input on scene creation in webui

### LINKED ISSUE #476: [Feature] Adding fingerprints manually
I would like the option of including other hashes manually from the UI. I think right now the only way to include fingerprints is by creating the scene from stash, but then the fingerprints are limited only to the ones that stash creates and it would be better to have more algorithms so that programs that use those algorithms may use the stash-box api without having to remake their hashing to include a new algorithm specifically to use for stash-box.

ikmckenz:
Mostly based this off the urlInput file. Tested working locally. 

Resolves #476, which I guess was originally about adding some new hash type but has been edited to just "add fingerprints manually", which is what this does.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ There are two ways to authenticate a user in Stash-box: a session or an API key.
 | `postgres.max_open_conns` | (0) | Maximum number of concurrent open connections to the database. |
 | `postgres.max_idle_conns` | (0) | Maximum number of concurrent idle database connections. |
 | `postgres.conn_max_lifetime` | (0) | Maximum lifetime in minutes before a connection is released. |
+| `require_scene_draft` | false | Whether to allow scene creation outside of draft submissions. |
 
 ## SSL (HTTPS)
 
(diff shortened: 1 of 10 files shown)


ikmckenz:
Rebased, and fixed formatting (yarn validate now passes)

DogmaDragon:
I strongly disagree with this being the default feature. While I can see a possible use case for this, without some safeguards it opens up a huge vector for abuse. 

While Stash requirement is not ideal, it automatically adds certain safeguards like:
- Hash being generated on actual file
- File duration being sent along with the hash
- Using the same algorithm as everyone else

ikmckenz:
Perhaps the entire “create scene” button in the webui should be configurable via a feature flag in that case? Operators who only want scenes via stash should disable entirely, and those operators who want it enabled will have users who can submit scenes with hashes in the webui.

ikmckenz:
I'll rework this and re-open from a different branch :)

PR #762, opened 2024-02-26
Remove show all from search result if there's no results

InfiniteStash:


frontend/src/components/searchField/SearchField.tsx
```diff
@@ -153,13 +153,18 @@ function handleResult(
...
-    ...(scenes.length ? [{ label: "Scenes", options: scenes }] : []),
-  ];
+  const performerResults = performers.length
+    ? [{ label: "Performers", options: performers }]
+    : [];
+  const sceneResults = scenes.length
+    ? [{ label: "Scenes", options: scenes }]
+    : [];
+  const showAll =
+    showAllLink && performerResults.length > 0 && sceneResults.length > 0
```
Flashy78, line 163: I think this should probably be:
`const showAll = showAllLink && (performerResults.length > 0 || sceneResults.length > 0)`

Right now if there are no performer results, but multiple scene results, I can't go to the main search results page to see which one I want, I have to pick one from the dropdown. If it's not the one I wanted, I have to re-type the search and choose another one from the dropdown, etc.

PR #798, opened 2024-07-23
Allow jpeg & jfif for Image Uploads

S3L3CT3DLoves:
Minor change, but will make life so much easier.

Uploading images (for performers, as an example) now allows using JPEG extension files (already supported, but missing in the filter)


DogmaDragon (dismissed):

BonerFide:
Not a problem with this change but .jfif could use the same treatment given how often chrome saves jpeg files with that extension.

S3L3CT3DLoves:
Added jfif to the list ! 👍

PR #800, opened 2024-07-23
Update session.go

L3GW0RK:
changed cookie name from "session" to "stashbox" to remove browser conflicts with stash instances (stashapp also uses "session"). this will allow users to be logged in to both their Stash app instance and StashBox/Stashdb at the same time in the same browser session.


InfiniteStash:
Good idea, thanks!

PR #803, opened 2024-08-09
Update docker compose for production

javstash:
Update docker-compose.yml and Dockerfile

* bump postgres to 15.8
* change the bktree repo from [fake-name/pg-spgist_hamming](https://github.com/fake-name/pg-spgist_hamming) to [evirma/pg_bktree](https://github.com/evirma/pg_bktree) and updated build instruction
* * The only difference between the two repo is the Makefile

* Change docker compose so now it mounts /traefik and then look for acme.json in /traefik
** avoid `Are you trying to mount a directory onto a file (or vice-versa)?` error

docker compose up should now run on Ubuntu 22.04 LTS + Docker without manual intervention


javstash:
closing it for now, will resubmit later

PR #824, opened 2024-10-15
Scene Drafts can now get multiple links

S3L3CT3DLoves:
This change allows Scene Drafts to be uploaded with multiple URLs.

**UX changes**
Unlike in the past, links that have no "Site" regex match will no longer be assigned the "Studio" site automatically. This was discussed on Discord. Instead, this replicates the current behaviour of the link parsing in Performer drafts.

**Breaking Changes**
This should not cause breaking changes, the "Urls" field is added to the GQL API. The old "Url" field should be deprecated, but this depends on Stash, so for now I intentionaly kept both fields available.

**Further Improvements**
I'll add another PR later to work on improving the way URLs are parsed in the Drafts frontend, to make sure URLs are mapped to Sites. But this will impact both Scenes and Performers, and is a different issue in my opinion.

This is my first time using Go, please let me know if I made some errors (although I didn't change much in the Go code tbh)

Resolves: https://github.com/stashapp/stash-box/issues/741

PR #825, opened 2024-10-15
Match URLs in PerformerDraft

S3L3CT3DLoves:
This will make the frontend parse the URLs which are passed to a Draft, and automatically assign them a Site and add them to the Draft's URLs.

It also keeps track of URLs that do not match any of the Site regexes, and keeps those in the existing "Remainder" zone, so the user can manually add those links if relevant.

This only affects PerformerDraft, but the code is generic so it will be easy to extend it to SceneDraft once  https://github.com/stashapp/stash-box/pull/824 is merged.


InfiniteStash:
Thanks, as suggested I've added it to the scene draft parsing as well.

PR #847, opened 2024-12-05
Change resize_linux.go to resize_unix.go

ikmckenz:
Also change the build flag. This fixes the build on MacOS and other unixes.


InfiniteStash:
Good call, thanks!

PR #858, opened 2024-12-11
Limit formats of uploaded images to jpeg/webp/svg

InfiniteStash:


DogmaDragon:
@InfiniteStash Could you extend this to .png? It's a popular format for studio images that doesn't have vectors. It's also sometimes used for performer images on some stash-box instances. Should have universal browser support.

PR #866, opened 2024-12-13
Add ability to sort performers by updated_at

### LINKED ISSUE #830: Possibility to sort performers with updated_at
The scenes can be sorted with the option "updated_at", it's very usefull for editing. This is not possible for the performers, can this option please be added? Thanks in advance!

ikmckenz:
Resolves #830 

Quick one line change since the backend infrastructure is already there. Not sure how performant the underlying query will be on a big stashbox, or if it needs an index, but it looks like the other sort options don't currently have indexes other than `name`.


InfiniteStash (approved):
There probably should be an index, but the impact is very small. Thanks!

Issue #870, opened 2024-12-21
[Feature] findUpdatedScenes


ikmckenz:
Opening up discussion on a `findUpdatedScenes` API, first described in stashapp/stash/issues/2903. Original author @mmenanno described it as follows: 
> ### Stash-Box Backend Side
> 
> For the sake of making the process optimal Stash-Box should have a new query `findUpdatedScenes` that takes an array of `SceneUpdateInput` structured something like:
> 
> ```json
> {
>   "scene_id": "stash_id",
>   "last_recorded_update": "external_updated_at"
> }
> ```
> 
> This call can return a `updated` Boolean, deleted Boolean, and `Scene` object for each input.
> 
>  * If the `last_recorded_update` is NOT older than the `updated_at` on the `scene_id`, it will return `updated: false` , `deleted: false`, and no scene object
> 
>  * If the `last_recorded_update` is older than the `updated_at` on the `scene_id` but the scene has been deleted, `updated: true`, `deleted: true`, and no scene object
> 
>  * If the `last_recorded_update` is older than the `updated_at` on the `scene_id` and the scene is not deleted, `updated: true`, `deleted: false`, and the updated Scene object
> 

To me maybe this seems to do too much on the stash-box backend, perhaps the input would be better as the list of stash_id's and the response could be the stash_id's and their last updated time, leaving the identification of more recently scenes to the local stash application. This also seems to mirror what @stg-annon was saying. Curious to hear people's thoughts.

stg-annon:
aside from publishing something like a list object ids and last modified timestamps, Id think the only thing we could maybe use on the sash-box side of things is a `findScenes` query that takes a list of IDs so you don't have to make a query for each scene to check if its been updated you could do it page by page

also adding a some kind of date modified field directly to objects that just takes the last accepted edit closed date and applies it to the object would help with not having to query a list of edits and just the main object and this field

ikmckenz:
> aside from publishing something like a list object ids and last modified timestamps

Just to put some pen to paper here I opened draft PR #871, which does this. But I also see the merits of just having the less-specific `findScenes` API as well.

InfiniteStash:
> > aside from publishing something like a list object ids and last modified timestamps
> 
> Just to put some pen to paper here I opened draft PR #871, which does this. But I also see the merits of just having the less-specific `findScenes` API as well.

[...]

PR #871, opened 2024-12-21
[DRAFT] Add ability to find scene updated times by id list

ikmckenz:
One potential solution for #870, specifically this is a new API that gives scene ids and updated_at timestamps from a query of a list of ids.


ikmckenz:
Build is failing but seems unrelated to my changes on first glance.

PR #876, opened 2024-12-29
Fix display issues for scenes without images

### LINKED ISSUE #853: [Feature] Add placeholder cover for scenes
Sometimes for one reason or another, scene can have no cover image.
In those cases scene card will show file can't be displayed error. See screenshot.

Adding an SVG placeholder for scenes with no cover image.

**Additional context**
image

### LINKED ISSUE #873: Scene page fails to load when there is no scene image
This is on develop.
A scene that has no image throws an exception, because `image` is `undefined`.
I guess the `style` attribute/`aspectRatio` needs to be conditional. (probably **not**  `` `${image?.width}/${image?.height}` ``)
https://github.com/stashapp/stash-box/blob/2a063aebfa9cc8eff48bf728385f9f1b8987cefc/frontend/src/components/image/Image.tsx#L86

Notes:
1. Performer: `ImageCarousel` uses `Image`, but has an additional check because of the sorting.
   https://github.com/stashapp/stash-box/blob/2a063aebfa9cc8eff48bf728385f9f1b8987cefc/frontend/src/components/imageCarousel/ImageCarousel.tsx#L24
3. Studio: Studio logo is not rendered using this component.
   https://github.com/stashapp/stash-box/blob/2a063aebfa9cc8eff48bf728385f9f1b8987cefc/frontend/src/pages/studios/Studio.tsx#L103

InfiniteStash:
Resolves #873

### DIFF
--- a/frontend/src/components/image/Image.tsx
+++ b/frontend/src/components/image/Image.tsx
@@ -13,7 +13,7 @@ type Image = {
 };
 
 interface ImageProps {
-  image: Image;
+  image?: Image;
   emptyMessage?: string;
   size?: number | "full";
   alt?: string;
@@ -29,7 +29,7 @@ const ImageComponent: FC<ImageProps> = ({
     "loading",
   );
 
-  if (!image.url)
+  if (!image?.url)
     return (
       <div className={`${CLASSNAME}-missing`}>
         <Icon icon={faXmark} color="var(--bs-gray-400)" />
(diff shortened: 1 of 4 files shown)


frontend/src/components/image/Image.tsx
```diff
@@ -80,11 +80,10 @@ const ImageContainer: FC<ContainerProps> = ({
     ? sortImageURLs(images, orientation)[0]
     : images;
 
+  const aspectRatio = image ? `${image.width}/${image.height}` : "16/6";
```
peolic, line 83: did you mean?
```suggestion
  const aspectRatio = image ? `${image.width}/${image.height}` : "16/9";
```

PR #894, opened 2025-01-07
Fix search page crash when search term contains slash

peolic:
The search page currently crashes when provided with a URL, which is needed for the [search-by-url feature](https://github.com/stashapp/stash-box/pull/843).

Uses https://reactrouter.com/6.8.1/route/route#splats
This looks a little weird with the asterisk, but it tells `react-router` to stop parsing the path, and just make the rest of it one single parameter, as is needed for the search page..


InfiniteStash (approved):
Thanks! I'd forgotten this was even implemented.

PR #904, opened 2025-01-10
Fix clearing fields in merge edits

### LINKED ISSUE #903: Clearing fields in merge edits
It appears to be the same bug as #587, which was fixed #677,
only dealt with `MODIFY` edits. `MERGE` edits still have this bug.
Example: Merging `Jayne` into `Jane [2025]` and removing the disambiguation results in an merge without the change.
Probably affects other merge-able entities.

It's possible that the fix #677 can be simply applied to the merge functions.

peolic:
Fixes #903, by applying the same fix to `MERGE` edits, as #677 did for `MODIFY` edits.

- [x] Tested only on performers merges, trying to remove disambiguation from the merge target.
- [ ] ~~Need to look into adding tests, like in #677?~~
		it seems I don't have any patience to do this

### DIFF
--- a/pkg/manager/edit/performer.go
+++ b/pkg/manager/edit/performer.go
@@ -123,7 +123,8 @@ func (m *PerformerEditProcessor) mergeEdit(input models.PerformerEditInput, inpu
 	}
 
 	// perform a diff against the input and the current object
-	performerEdit, err := input.Details.PerformerEditFromMerge(*performer, mergeSources, inputArgs)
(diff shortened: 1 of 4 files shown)


InfiniteStash (approved):
I don't blame you for not writing tests, it's entirely too much boilerplate at the moment. I need to rewrite them to be faster to set up, but dealing with tests is the most boring thing in the world...

Anyway, thanks!

PR #957, opened 2025-06-02
Add optional ignore_size to image resize settings

ikmckenz:
This optional setting exempts images under ignore_size from being resized, saving on server CPU.

Not sure if this is fully necessary after the recent changes to add a resize cache, but if there are further CPU usage concerns from resizing then this should help. Defaults to 0, which means when not set all images will be resized as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,7 @@ There are two ways to authenticate a user in Stash-box: a session or an API key.
 | `image_max_size` | (none) | Max size of image, if no size is specified. Omit to return full size. |
 | `image_resizing.enabled` | false | Whether to resize images shown in the frontend. |
 | `image_resizing.cache_path` | (none) | Folder in which resized images will be saved for later requests. Recommended when resizing is enabled. |
+| `image_resizing.min_size` | (none) | Only resize images above a certain size |
 | `userLogFile` | (none) | Path to the user log file, which logs user operations. If not set, then these will be output to stderr. |
 | `s3.endpoint` | (none) | Hostname to s3 endpoint used for image storage. |
 | `s3.bucket` | (none) | Name of S3 bucket used to store images. |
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -92,6 +92,7 @@ There are two ways to authenticate a user in Stash-box: a session or an API key.
 | `image_max_size` | (none) | Max size of image, if no size is specified. Omit to return full size. |
 | `image_resizing.enabled` | false | Whether to resize images shown in the frontend. |
 | `image_resizing.cache_path` | (none) | Folder in which resized images will be saved for later requests. Recommended when resizing is enabled. |
+| `image_resizing.ignore_size` | (none) | Only resize images above a certain size |
```
InfiniteStash: I think `min_size` is maybe a bit clearer.

pkg/api/routes_image.go
```diff
@@ -80,7 +80,8 @@ func (rs imageRoutes) image(w http.ResponseWriter, r *http.Request) {
 	w.Header().Add("Cache-Control", "max-age=604800000")
 
 	// Resize image
-	if maxSize != 0 && config.GetImageResizeConfig().Enabled {
+	ignoreSize := config.GetImageResizeConfig().IgnoreSize
+	if maxSize != 0 && config.GetImageResizeConfig().Enabled && (databaseImage.Width > int64(ignoreSize) || databaseImage.Height > int64(ignoreSize)) {
 		if databaseImage.Width > int64(maxSize) || databaseImage.Height > int64(maxSize) {
```
InfiniteStash: Can you break these two conditionals out into a `shouldResize(...)` function? It's pretty hard to read atm.

ikmckenz: Added in new version

README.md
ikmckenz: Added in new version

PR #961, opened 2025-06-05
Set security headers

InfiniteStash:


feederbox826:
Any specific reason why CSP was added? This breaks multiple userscripts including backlog

PR #962, opened 2025-06-05
Only set CSP necessary to fix security issue

feederbox826:


feederbox826:
Superceded by #963, fixes the core problem instead of monkeypatch with CSP

PR #963, opened 2025-06-05
Disallow markdown in ReactMarkdown to prevent XSS leak

feederbox826:
supercedes #962, fixes the issue at it's core


feederbox826:
Implemented in https://github.com/stashapp/stash-box/pull/964

PR #967, opened 2025-06-08
Treat unmatched URLs as studio even if more than one URL was sent

smith113-p:
Now that Stash supports sending more than one URL at a time (https://github.com/stashapp/stash/pull/5894), this code treats unmatched URLs as a studio URL even if multiple URLs were submitted.

This allows, for example, submitting a Clips4Sale URL and a studio URL and having both matched.


smith113-p:
Note: I have not tested this yet because I do not have postgresql rn

smith113-p:
Thanks to Infinite I was able to test this, and it works correctly:
Screenshot 2025-06-08 175452

I have also tested that a single studio URL and a single non-studio URL still work correctly.

Issue #927, opened 2025-01-29
Remove EDIT role requirement from notification subscriptions


jaoxford:
Hey,

I think we should be able to receive notifications when new scenes from studios / performers are released.

jaoxford:
I believe changing this line:

https://github.com/stashapp/stash-box/blob/0ce51709e4addb1ae2cbd9e51e82619a4fa6d947/graphql/schema/schema.graphql#L193

Could do it, but would there be any side effects or other changes that would need to be done?

peolic:
Click the notification bell, then "Edit Subscriptions". You can subscribe for both:
- [x] A new scene from a studio you have favorited.
- [x] A new scene involving a performer you have favorited.

jaoxford:
Whilst this is correct, I had to receive the `EDIT` role from an admin. I think this should be useable by all users.

DogmaDragon:
It would make sense to remove the role requirement for favorite-type notifications, but the requirement makes sense for the rest.

PR #982, opened 2025-09-16
Allow "Read" Role to Edit Favorite Notifications

sashapp:
As mentioned in issue #927, normal users cannot set notifications for their favorite studios and performers. This PR remedies the issue by adding a second API call `updateFavoriteNotificationSubscriptions` requiring only the `READ` role. In the backend, a check is performed, such that indeed only the favorite notifications can be edited (other notifications are filtered out and cannot be set via the call).


DogmaDragon:
They are not "admin" notifications. They require EDIT role to use. Using admin terminology makes no sense.

sashapp:
You are right. I changed it from admin to editor. Do you like this better or do you have another proposal?

sashapp:
I fixed the failing pipeline.

sashapp:
@DogmaDragon Do you agree with the renaming from admin to editor subscriptions?

DogmaDragon:
That sounds better to me, but @InfiniteStash has the final say.

sashapp:
@InfiniteStash what are your thoughts on this PR?

InfiniteStash:
> @InfiniteStash what are your thoughts on this PR?

Hey, thanks for the PR.

I would prefer just using the existing mutation and changing it to `READ` permission. Having multiple mutations for the same thing is more work to maintain. We can change the mutation to filter the notifications on the backend and only persist favorites notifications for anyone without edit perms, and similarly only show favorite notifications on the frontend for anyone missing edit.

sashapp:
Perfect, will do :)

sashapp:
@InfiniteStash @DogmaDragon I have created a second proposal in [Allow "Read" Role to Edit Favorite Notifications](https://github.com/stashapp/stash-box/pull/992).

PR #1000, opened 2025-11-20
Fix otelpgx span names and add username to graphql errors

InfiniteStash:


discourse-stashapp:

PR #1015, opened 2025-12-04
Fix null age for partial birthdates

### LINKED ISSUE #1001: [Bug Report] GraphQL queryPerformers returns empty age
**Describe the bug**
GraphQL `queryPerformers` queries that query for a specific age can return performers with an age of null.

**To Reproduce**
1. Run the following query against stashdb.org in your favorite GraphQL explorer
```graphql
{
  queryPerformers(input: {
    name: "Marley Kunis",
    age: {value: 31, modifier: EQUALS}
  }) {
    count
    performers {
      id
      age
    }
  }
}
```
**Expected behavior**
When querying for a specific age, I expect all performers to be of that age.

**Additional context**
This is a discrepancy between how the SQL query [here](https://github.com/stashapp/stash-box/blob/883a1500f05c298fc0b79a9181acc13ccfc83926/internal/service/performer/query.go#L111), and the resolver logic [here](https://github.com/stashapp/stash-box/blob/883a1500f05c298fc0b79a9181acc13ccfc83926/internal/api/resolver_model_performer.go#L42), handle incomplete dates. Postgres accepts incomplete dates and sets all the missing parts to `01` , but the resolver tries to parse the date and bails when that fails.

One solution would be to 'fill up' the dates if they are too short, that is if a date is only a year, change the date passed to `ParseDateStringAsTime` to `xxxx-01-01`, and similarly for dates that are only a year and a month (maybe this can be done inside `ParseDateStringAsTime`).

InfiniteStash:
Resolves #1001

### DIFF
--- a/pkg/utils/date.go
+++ b/pkg/utils/date.go
@@ -5,47 +5,20 @@ import (
 	"time"
 )
 
-const railsTimeLayout = "2006-01-02 15:04:05 MST"
-
-// PadIncompleteDate pads incomplete date strings (YYYY or YYYY-MM) to complete dates (YYYY-MM-DD)
-func PadIncompleteDate(date string) string {
-	switch len(date) {
-	case 4:
-		return fmt.Sprintf("%s-01-01", date)
-	case 7:
-		return fmt.Sprintf("%s-01", date)
-	default:
-		return date
-	}
-}
-
 func ParseDateStringAsTime(dateString string) (time.Time, error) {
-	// https://stackoverflow.com/a/20234207 WTF?
-
-	t, e := time.Parse(time.RFC3339, dateString)
-	if e == nil {
-		return t, nil
-	}
-
-	t, e = time.Parse("2006-01-02", dateString)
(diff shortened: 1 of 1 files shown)


JustusJ:
I know I suggested the solution used here, but I just saw that Stash had a similar problem [here](https://github.com/stashapp/stash/pull/6305/files). They use `time.Parse("2006-01", dateString)` and `time.Parse("2006", dateString)`, which I think could be used here as well. I think that that is a clean solution, and maybe it makes sense to keep them aligned.

InfiniteStash:
> I know I suggested the solution used here, but I just saw that Stash had a similar problem [here](https://github.com/stashapp/stash/pull/6305/files). They use `time.Parse("2006-01", dateString)` and `time.Parse("2006", dateString)`, which I think could be used here as well. I think that that is a clean solution, and maybe it makes sense to keep them aligned.

Thanks, I've simplified it a bit.

discourse-stashapp:

PR #1016, opened 2025-12-04
Fix performers with 0 scenes being dropped from queries

### LINKED ISSUE #1002: [Bug Report] GraphQL queryPerformers: Sorting by scene count omits performers with 0 scenes
**Describe the bug**
GraphQL `queryPerformers` queries that are sorted by `SCENE_COUNT` do not return any performers that have 0 scenes. This also leads to differences between the `count` field and the actual count of performers in the result.
The same is true when sorting by `DEBUT` or `LAST_SCENE`.

**To Reproduce**
1. Run the following query against stashdb.org in your favorite GraphQL explorer (that particular performer happens to have 0 scenes at the moment)
```graphql
{
  queryPerformers(input: {
    sort: SCENE_COUNT,
    name: "hidden_foxy"
  }) {
    count
    performers {
      scene_count
    }
  }
}
```

**Expected behavior**
Performers with 0 scenes are returned, and the `count` field and the number of performers in the result agree.

**Additional context**
I think this is because of the Inner Join [here](https://github.com/stashapp/stash-box/blob/883a1500f05c298fc0b79a9181acc13ccfc83926/internal/service/performer/query.go#L184), which should be a `LEFT JOIN` to also include performers that have no scene.

Note that the Join might already have been applied to the query before this point, so fixing it only here might not be enough. I'm not sure what the consequences are of changing those queries to `LEFT JOIN`s.

InfiniteStash:
Resolves #1002

### DIFF
--- a/internal/api/performer_integration_test.go
+++ b/internal/api/performer_integration_test.go
@@ -660,6 +660,149 @@ func (s *performerTestRunner) testQueryPerformersByAgeAndBirthYear() {
 	assert.True(s.t, found4, "Performer currently age 20 not found")
 }
 
+func (s *performerTestRunner) testQueryPerformersSceneCountSort() {
+	// Test that performers with 0 scenes are returned when sorting by SCENE_COUNT
+	// Create performer with no scenes
+	performerWithNoScenes, err := s.createTestPerformer(&models.PerformerCreateInput{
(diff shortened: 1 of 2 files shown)


discourse-stashapp:

PR #1027, opened 2025-12-21
Fix multi criterion query plan

InfiniteStash:
`WHERE IN` causes postgres to read the whole primary table before joining, which is very slow. `INNER JOIN` lets it start with the subquery instead.


discourse-stashapp:

https://discourse.stashapp.cc/t/release-v0-7-8-stashapp-stash-box/4867/1

PR #1030, opened 2026-01-06
Fix safari image sizing in edit queue and image upload

### LINKED ISSUE #931: [Bug Report] Images in edits doesn't have correct aspect ratio
**Describe the bug**
Images doesn't have correct aspect ratio in Safari (Mac & iOS). Applies to all images.

Image

**OS:**
 - OS: Mac & iOS
 - Browser: Safari
 - Version: iOS 18 & Sonoma

Placeholder "Loading image" doesn't show either.

gimmeliina:
Closes #931. There are be many more similar issues here and there but this should fix the most important ones

### DIFF
--- a/frontend/src/components/imageChangeRow/styles.scss
+++ b/frontend/src/components/imageChangeRow/styles.scss
@@ -2,8 +2,9 @@
   display: flex;
   flex-wrap: wrap;
 
-  .Image {
+  &-image {
     height: 150px;
     margin: 5px;
+    width: auto;
   }
 }
--- a/frontend/src/components/imageChangeRow/styles.scss
+++ b/frontend/src/components/imageChangeRow/styles.scss
@@ -2,7 +2,7 @@
   display: flex;
   flex-wrap: wrap;
 
-  &-image {
+  .Image {
     height: 150px;
     margin: 5px;
     width: auto;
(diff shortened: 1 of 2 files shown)


frontend/src/components/imageChangeRow/styles.scss
gimmeliina, line 1: This is not fixed. In my testing it worked, but changing ` &-image` to `. Image` was the magic trick. Didn't found the root cause, but I remember that double definition of size messed up safari rendering. I will look this again if I find root cause; Based on [stackoverflow thread](https://stackoverflow.com/questions/13237782/safari-doesnt-calculate-rem-units-correct-when-scaling-with-media-width-heigh) linked in issue, it might be an issue with aspect ratio calculations.

gimmeliina:
Digging this little more and adding `aspect-ratio: unset !important` to `.Image` solves problem entirely. Is there any practical meaning to add it in style tag? 

https://github.com/stashapp/stash-box/blob/3f74bf2060b10ad13315fee2a79b076115db3e58/frontend/src/components/image/Image.tsx#L88C47-L88C70
#667 @InfiniteStash 

Related:
https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/

PR #1045, opened 2026-02-16
Change varchar columns to text

### LINKED ISSUE #944: [Bug Report] Scene creation does not check for size limit
**Describe the bug**
There seems to be a size limit of 255 characters for the title (see the automated comment on https://stashdb.org/edits/48db021f-7c4e-41d8-a797-2206de3eccc6. Note the scene image/text is NSFW)

But I was not aware of the limit. The limit is only checked once the edit has enough upvotes and the edit fails.

**To Reproduce**
1. Submit a "Create a scene" edit with a long title
2. Wait for enough upvotes
3. You get a notification that the edit has failed. If you click on the edit, you can see an automated comment:
```
Edit application failed:
Unknown Error: Error creating Scene: pq: value too long for type character varying(255)
```

**Expected behavior**
If a proposed edit exceeds some limit, I should get an error when creating the edit, not later on when creation fails

**Screenshots**
n/a

**Desktop (please complete the following information):**
 - OS: Probably Windows. Might have been Android
 - Browser: Chrome
 - Version: 137

**Additional context**
A couple of other things I noticed:
- The failed edit only shows 2 upvotes. Presumably, there was a third upvote that triggered the attempt to create the scene, but because the SQL command failed, the third vote was not recorded (?)
- The notification only said "edit failed". The automated comment did not result in a notification, so I did not know that details were available at all until I clicked through to the edit itself.

InfiniteStash:
There is no practical difference between varchar/text in postgres, except varchar has a restricted length.

The length restriction can cause edits to fail, so I'm changing any columns that can conceivably be longer than 255 characters to text.


smith113-p:
This effectively closes issue #944

PR #1085, opened 2026-04-27
feat: implement modern multi-stage Dockerfile using Go 1.25 and pnpm

markd3ng:
## Summary
The current build Dockerfile is outdated (using Go 1.13 and Ubuntu 19.10). This PR introduces a modern, multi-stage Dockerfile to improve build speed, security, and image size.

## Changes
- Created a new multi-stage `Dockerfile` in `docker/production/stash-box/`.
- Upgraded build environment to **Go 1.25** and **Node 24**.
- Switched from yarn/npm to **pnpm** for frontend builds to leverage better caching and speed.
- Used a lean **Ubuntu 24.04** runtime image, significantly reducing the final image footprint.
- Separated frontend and backend build stages for cleaner layer management.

## Why
Modernizing the build pipeline ensures compatibility with the latest Go features and provides a faster, more reliable deployment path for production environments.

### DIFF
--- a/docker/production/stash-box/Dockerfile
+++ b/docker/production/stash-box/Dockerfile
@@ -1,0 +1,43 @@
+# Multi-stage build for stash-box
+# Must be built from the project root:
+#   docker build -f docker/production/stash-box/Dockerfile .
+
+# Stage 1: Build frontend
+FROM node:24-slim AS frontend
+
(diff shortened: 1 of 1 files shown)


InfiniteStash:
We do not have dockerfiles for production because nobody should be building the binary for production.

PR #1086, opened 2026-04-27
New: Move Vote Bar Below Comments

Gykes:
This was initially noticed in Discourse, https://discourse.stashapp.cc/t/move-voting-controls-below-comments-to-encourage-informed-voting/6961 

This was a small addition that was brought up to me in hopes of helping reduce voters from just automatically clicking +1. 

UI: 

Screenshot 2026-04-27 at 14 49 07
Screenshot 2026-04-27 at 14 48 10

### DIFF
--- a/frontend/src/components/editCard/EditCard.tsx
+++ b/frontend/src/components/editCard/EditCard.tsx
@@ -94,7 +94,6 @@ const EditCardComponent: FC<Props> = ({
             <b className="me-2">Status:</b>
             <EditStatus {...edit} />
             <EditExpiration edit={edit} />
-            {showVoteBar && <VoteBar edit={edit} />}
           </div>
         </div>
       </Card.Header>
@@ -110,6 +109,7 @@ const EditCardComponent: FC<Props> = ({
                 {showVotes && <Votes edit={edit} />}
                 {comments}
                 <AddComment editID={edit.id} />
+                {showVoteBar && <VoteBar edit={edit} />}
               </Col>
             </Row>
           </>


discourse-stashapp:

https://discourse.stashapp.cc/t/move-voting-controls-below-comments-to-encourage-informed-voting/6961/4

DogmaDragon:
I can see the reasoning, but I think the implementation is bad. 

`Add Comment` button is next to the votes now which will inevitably lead misclicks.
This also forces you to scroll up and down to see other voting details and with edits that have multiple comments it becomes unwieldy quite fast. 
E.g. https://stashdb.org/edits/019dbac2-8221-7022-b85c-19a4c03d6cba

TPG1000:
Another consideration is some of us review the edit queue from last to first. I start at the bottom of the page and work up. whilst the change is a good thought I'd have to scroll up & then back down to vote.

Gykes:
The generalized thought process was that you should have to scroll all the way down so that someone can't say they didn't see something. Sure, it may add 1% more work to people who actually ready everything and vote but now people can't play dumb when they just upvote without looking. 

My first thought is terms of service. The "I Agree" button isn't at the top, it's at the bottom and you gotta scroll for 3 business days to get there. The reason it's that way is so that users cant play dumb and say they never read it. You had to scroll and click agree.... it's on them at that point. 

> Add Comment button is next to the votes now which will inevitably lead misclicks.

The Add Comment button could be moved to the left side of the box instead of the right to help mitigate this

> This also forces you to scroll up and down to see other voting details and with edits that have multiple comments it becomes unwieldy quite fast.

Agreed but this is kind of the point. It shouldn't be so easy to just click the button and move on with no consequences for not doing your diligence 

EDIT as I forgot to reply to it.

> Another consideration is some of us review the edit queue from last to first. I start at the bottom of the page and work up. whilst the change is a good thought I'd have to scroll up & then back down to vote.

This I think is a problem that can be solved by sorting the edit list oldest to newest rather than newest to oldest. Could potentially be another PR. Seems odd that you need to go to the last page to get to the "front" of the queue.

feederbox826:
While this is an overall good change, I don't think this is a fitting change nor does it address the  core issue we face. Needs a bit more discussion and probably a different implementation

DogmaDragon:
> Agreed but this is kind of the point. It shouldn't be so easy to just click the button and move on with no consequences for not doing your diligence

I don't think that adding intentional friction for everyone, including vigilant editors is a good solution to cut down on bad behaviour. It will just frustrate everyone without making much difference. 

I agree that that UI and the workflow could be improved, but I see bad voting as an intentional action of "I don't care enough about this" instead of an honest oversight of someone not seeing the ongoing discussion in the comments.

Maista6969:
I'm with Dogma on this: I don't think a simple technical change can fix what is fundamentally a social/political problem

PR #1089, opened 2026-05-04
Change performer image prioritization; use float division for aspect ratios

tj2222:
### Bug(?) fix: Use float division for aspect ratios
* When sorting landscape (scene) images and portrait (performer) images, integer division was being used to determine aspect ratio. This meant that, e.g., a 4:3 image (1.333) and a 16:9 image (1.777) would be considered to have equal aspect ratios, leaving the "tie" to be broken by prioritizing the image with wider pixel dimensions.
* I *believe* this was a bug, but if it's the intended behavior, that ought to be documented.

### Feature change: Performer image prioritization
* Prioritizes performer images by closeness to [StashDB Guidelines'](https://guidelines.stashdb.org/docs/performers/edit/performer-images/performer-aspect-ratio/) ideal 2:3 aspect ratio, by modifying OrderPortrait (performer images are the only use of this function).
  * I thought I'd heard that this was already the behavior, but that probably either regressed or was something I imagined.
  * Perhaps this isn't the ideal ordering, but it's subjective and seems reasonable to me. Open to alternatives!
* Inspired by my frustration with [this performer](https://stashdb.org/performers/019de86e-b937-7d17-9217-313531edd116)'s images, where I cropped the only face photo to 2:3 aspect ratio, but a photo without the performer's full face was shown as the first image.

## Notes
* I'm very new to golang, this is my first time working with the stash-box code, and this code was written with assistance from Antigravity/Gemini 3.1 Pro (though I do understand it, I think, and wrote a lot of it "by hand"). This is all to say: Please take a careful look to see if there's anything dumb in here.
* I have not tested this with a full stash-box instance, only with unit tests.

### DIFF
--- a/internal/image/sort.go
+++ b/internal/image/sort.go
@@ -7,13 +7,14 @@ import (
 	"github.com/stashapp/stash-box/internal/models"
 )
 
+// Sorts by "most" to "least" landscape, i.e. largest to smallest aspect ratio; ties broken by largest --> smallest width.
 func OrderLandscape(p []models.Image) {
 	sort.Slice(p, func(a, b int) bool {
 		if p[a].Height == 0 || p[b].Height == 0 {
 			return false
 		}
-		aspectA := p[a].Width / p[a].Height
-		aspectB := p[b].Width / p[b].Height
+		aspectA := float64(p[a].Width) / float64(p[a].Height)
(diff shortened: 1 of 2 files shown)


InfiniteStash:
I'm pretty sure this used to work correctly, but hard to be sure without tests. Anyway, looks good, thanks!

Issue #930, opened 2025-02-13
[Feature] Sort performer scenes based on scene length


DontHugTheDishwasher:
**Is your feature request related to a problem? Please describe.**
When I have a scene locally but don't know what it's called in StashDB I usually find the performer and try to find the scene which matches the thumbnail and duration, but because it is currently not possible to sort based on duration this can take a very long time. If I were able to sort the scenes based on duration this would greatly reduce the time searching.

tj2222:
~~Sorting by duration is now available on the performer scenes list, so I think this can be closed.~~

Edit: I mistakenly thought this was `stash`, not `stash-box`. 🤦
Issue should remain open.

DogmaDragon:
@tj2222 where do you see an option to sort by duration?

tj2222:
@DogmaDragon D'oh. Thought this was Stash. Edited earlier comment.

PR #1092, opened 2026-05-15
Add support for sorting scenes by duration

tj2222:
Implements feature request from #930 to support sorting a performer's scenes by duration, as well as in other scene list views.

Tested manually on a local `stash-box` instance, and with an [integration test](https://github.com/stashapp/stash-box/commit/0e0553c3bd32573c1d3409bf2ae91562e18396b5). I've chosen to not include that test in this PR under the questionable logic of "if sorting was something that ought to be tested, there would already be tests for it". If it's appropriate to test scene sorting, I'd be happy to add tests for all sort options.

LLM use (Antigravity/Gemini): Implemented "by hand"... kind of. Used LLM-based tab completion. Above test was written entirely by LLM.


discourse-stashapp:

https://discourse.stashapp.cc/t/sort-filter-by-scene-length/4803/2

PR #1101, opened 2026-05-20
Fix queryScenes fingerprint criterion

InfiniteStash:
Fingerprint criterion is still string, so we need to convert to int64 before hitting sql.


discourse-stashapp:

https://discourse.stashapp.cc/t/release-v0-8-1-stashapp-stash-box/7356/1

PR #1103, opened 2026-05-20
Tweak performer alias search scoring

InfiniteStash:
Disables length normalization and frequency tracking for performer aliases, so performers with many aliases aren't penalized in search scoring, or boosted if they have a lot of aliases matching the search term.


discourse-stashapp:

https://discourse.stashapp.cc/t/release-v0-8-1-stashapp-stash-box/7356/1

PR #1111, opened 2026-05-22
Improve scene search ranking

InfiniteStash:
Drafted with claude.

Improves scene search by:
* Adding release date to the index
* Adding studio_aliases, network_aliases
* Rebuilds the index with `fieldnorms=false, record="basic"` for `performer_names`, `studio_aliases`, `network_aliases`. This means the search matches values in those arrays without penalizing long arrays. Previously it would prefer solo scenes over multi-performer scenes.
* Boosts performer names by x2 to avoid getting crowded out by scenes that contain the performer name in title/studio.

Also adds a first attempt at regression tests for the full text search.


discourse-stashapp:

https://discourse.stashapp.cc/t/better-search/1350/5

PR #1113, opened 2026-05-25
Add ordering scene/performers by popularity

### LINKED ISSUE #786: [Feature] Sort by total number of submissions/fingerprints
Users can use stash-box instances to browse for content by studio/performer etc.
To improve the experience another sort option similar to "Trending" can be added.
Sort by "Number of submissions" or "Popularity"

The new sort option should order scenes by the total number of fingerprint submissions.
Number of submissions for all different hashes should be added up. 

Attempting to use Trending produces no results for less recently active studios/performers.

**Additional context**
None

InfiniteStash:
Drafted with claude.

Adds materialized views for unique user submissions per scene/performer. That way we can order by number of people having submitted a scene/performer fingerprint, as an analog of the popularity of a scene or performer.

It works so-so for performers since performers with a lot of scenes will crowd out smaller performers, but it's still interesting. For scenes it works well.

The matview also allows us to remove the 3GB fingerprint created_at index we previously used for trending. The scene matview is 127MB for comparison, at the cost of being refreshed hourly rather than live.


discourse-stashapp:

https://discourse.stashapp.cc/t/sort-by-popular/1842/9

PR #1117, opened 2026-05-28
Remove fingerprints/images from deleted scenes

InfiniteStash:
Fingerprint/image deletion was accidentally broken in the big refactor. Reinstated, along with a migration to delete any leftovers.

### DIFF
--- a/internal/service/edit/scene.go
+++ b/internal/service/edit/scene.go
@@ -421,7 +421,15 @@ func (m *SceneEditProcessor) applyDestroy(scene *models.Scene) error {
 		return err
 	}
 
-	return err
+	if err = m.queries.DeleteSceneImages(m.context, scene.ID); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


internal/service/edit/scene.go
```diff
@@ -421,7 +421,11 @@ func (m *SceneEditProcessor) applyDestroy(scene *models.Scene) error {
 		return err
 	}
 
-	return err
+	if err = m.queries.DeleteSceneImages(m.context, scene.ID); err != nil {
+		return err
```
smith113-p, line 425: feels like we should try to delete the fingerprints even if deleting the images failed?

InfiniteStash: It's all running in a transaction, so if one of the queries fail, they're all rolled back.

You sometimes see it where edits fail to apply because of preconditions or unique contraints. The edit is failed and it posts an edit note about the failure.

PR #1123, opened 2026-05-29
Block scene/performer submits with pending URLs

### LINKED ISSUE #772: [Feature] Validate user hasn't forgotten to click 'add' for a link
I've noticed that many times I and other users have reported pasting in a link, and forgetting to click 'add' before submitting a studio/scene etc. This is only usually noticed when it was a mandatory link someone else picks up is missing. 

Possibly if you have put something in the URL field on the links tab, and you click 'Next' you should get a validation error asking to add the link or remove the link.

Alternatively it could be automatically added on navigate. 
(You will still get a chance to confirm on the confirm page. I think more people are missing links by not hitting 'add' than would be at risk of accidentally adding an extra link, which is more noticeable than a missing link. 
This option may not work in cases where the link does not match an regex to select the type of link. (If the stashbox allows arbitrary links that may not be a problem?)

smith113-p:
When I started contributing to stashdb, I constantly forgot to click the "Add" button when adding a URL to a scene draft.

This PR adds checks to the scene and performer draft forms to ensure that the user actually clicks Add after entering a URL.

Screenshot:
Screenshot 2026-05-29 135225

Drafted with Codex, but I have looked through the code myself and it looks reasonable to me (although my React knowledge is limited)

### DIFF
--- a/frontend/src/components/urlInput/urlInput.tsx
+++ b/frontend/src/components/urlInput/urlInput.tsx
@@ -31,9 +31,17 @@ interface URLInputProps {
   lens: Lens<URLItem[]>;
   type: ValidSiteTypeEnum;
   errors?: ErrorsType;
+  pendingURLError?: string;
+  onPendingURLChange?: (hasPendingURL: boolean) => void;
 }
 
-const URLInput: FC<URLInputProps> = ({ lens, type, errors }) => {
+const URLInput: FC<URLInputProps> = ({
+  lens,
+  type,
+  errors,
+  pendingURLError,
+  onPendingURLChange,
+}) => {
   const interop = lens.interop();
   const {
     fields: urls,
@@ -70,6 +78,7 @@ const URLInput: FC<URLInputProps> = ({ lens, type, errors }) => {
     if (inputRef.current) inputRef.current.value = "";
     setSelectedSite(undefined);
     setNewURL("");
+    onPendingURLChange?.(false);
   };
 
   const handleInput = (url: string) => {
@@ -159,14 +168,19 @@ const URLInput: FC<URLInputProps> = ({ lens, type, errors }) => {
           ref={inputRef}
           onBlur={(e) => handleInput(e.currentTarget.value)}
           placeholder="URL"
-          onChange={(e) => setNewURL(e.currentTarget.value)}
+          onChange={(e) => {
+            const value = e.currentTarget.value;
+            setNewURL(value);
+            onPendingURLChange?.(value.trim().length > 0);
+          }}
           onPaste={handlePaste}
-          className="w-50"
+          className={`w-50 ${pendingURLError ? "is-invalid" : ""}`}
         />
         <Button onClick={handleAdd} disabled={!newURL || !selectedSite}>
           Add
         </Button>
       </InputGroup>
+      {pendingURLError && <div className="text-danger">{pendingURLError}</div>}
     </div>
   );
 };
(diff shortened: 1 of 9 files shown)


frontend/src/components/urlInput/urlInput.tsx
```diff
@@ -159,14 +168,19 @@ const URLInput: FC<URLInputProps> = ({ lens, type, errors }) => {
           ref={inputRef}
           onBlur={(e) => handleInput(e.currentTarget.value)}
           placeholder="URL"
-          onChange={(e) => setNewURL(e.currentTarget.value)}
+          onChange={(e) => {
+            const value = e.currentTarget.value;
+            setNewURL(value);
+            onPendingURLChange?.(value.trim().length > 0);
+          }}
```
Gykes: If im reading this correctly the onPendingURLChange notification is wired into onChange. It looks like the existing onPaste handler also calls setNewURL directly if it matches a site. I think if a user pastes in a URL, doesnt add, then saves it will get missed.

frontend/src/pages/performers/performerForm/PerformerForm.tsx
```diff
@@ -668,6 +670,12 @@ const PerformerForm: FC<PerformerProps> = ({
             lens={lens.focus("urls").defined()}
             type={ValidSiteTypeEnum.PERFORMER}
             errors={errors.urls}
+            pendingURLError={errors.pendingUrl?.message}
+            onPendingURLChange={(hasPendingURL) =>
+              setValue("pendingUrl", hasPendingURL ? "pending" : "", {
```
Gykes: Soring pending vs "" to drive a boolean check seems weird to me and potentially a smell down the line. I would make pendingURL a yup.boolean() or store the pending url.

Gykes: Seems like this is also the same on SceneForm.tsx and schema.ts

frontend/src/pages/scenes/sceneForm/SceneForm.tsx
```diff
@@ -500,6 +503,12 @@ const SceneForm: FC<SceneProps> = ({
             lens={lens.focus("urls").defined()}
             type={ValidSiteTypeEnum.SCENE}
             errors={errors.urls}
+            pendingURLError={errors.pendingUrl?.message}
+            onPendingURLChange={(hasPendingURL) =>
+              setValue("pendingUrl", hasPendingURL ? "pending" : "", {
+                shouldValidate: true,
+              })
+            }
```
Gykes: Seems like this is duplicated. My genera rule of thumb is 2+ times to break it out into a hook or a shared component to reduce LOC

Gykes (changes requested):
For some reason I have lost the ability to wrap text in blocks. I beleive my key is broken.... sorry.

Just a quick once over. Didn't dig too deep into it.

frontend/src/components/urlInput/urlInput.tsx
smith113-p: Refactored this, thanks for the suggestion, the new code is a lot cleaner I think.

frontend/src/pages/performers/performerForm/PerformerForm.tsx
smith113-p: Made it store the pending URL.

frontend/src/pages/scenes/sceneForm/SceneForm.tsx
smith113-p: Done.

smith113-p (commented):
Please take another look!

PR #1128, opened 2026-05-31
Fix slow cluster queries

InfiniteStash:
Cluster queries were slow on prod since the scene_fingerprints table was misanalyzed, and config nuances meant the planner didn't use the correct index.

### DIFF
--- a/internal/database/database.go
+++ b/internal/database/database.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	postgresDriver = "postgres"
-	schemaVersion  = 63
+	schemaVersion  = 64
 )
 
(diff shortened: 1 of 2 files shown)


Maista6969:
Also noticed that these are available even with just the EDIT role, do we want to limit it to MODERATOR?

InfiniteStash:
> Also noticed that these are available even with just the EDIT role, do we want to limit it to MODERATOR?

It's intentional, I figure it's interesting to check for errors for non-mods as well, even if they can't move stuff. I've limited distances to max 8 for editors, but maybe I'll lower it another step.

Either way, it's way faster now, so not a big concern I think.

PR #1149, opened 2026-06-04
Query optimizations

InfiniteStash:
Optimizes a few different queries that show up in traces.
Also deprecates the `text` filter and removes `description` searching. It's very slow since it's unindexed, and and index would be enormous due to the corpus size. As far as I know nobody actually uses it for desc searching.

### DIFF
--- a/internal/service/scene/query.go
+++ b/internal/service/scene/query.go
@@ -106,13 +106,9 @@ func (s *Scene) buildSceneQuery(psql sq.StatementBuilderType, input models.Scene
 		) SFP ON scenes.id = SFP.scene_id`, userID)
 	}
 
-	// Filter by text (title and details)
 	if input.Text != nil && *input.Text != "" {
 		searchTerm := "%" + *input.Text + "%"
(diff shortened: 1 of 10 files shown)


internal/service/scene/query.go
```diff
@@ -106,13 +106,9 @@ func (s *Scene) buildSceneQuery(psql sq.StatementBuilderType, input models.Scene
 		) SFP ON scenes.id = SFP.scene_id`, userID)
 	}
 
-	// Filter by text (title and details)
 	if input.Text != nil && *input.Text != "" {
 		searchTerm := "%" + *input.Text + "%"
-		query = query.Where(sq.Or{
-			sq.ILike{"scenes.title": searchTerm},
-			sq.ILike{"scenes.details": searchTerm},
```
peolic, line 114: > As far as I know nobody actually uses it for desc searching.

I previously used it to find scenes with HTML tags in their descriptions, 
but that's not really a good use case anyway, even more so if it pollutes the database.

PR #1150, opened 2026-06-04
Fix image size calculation

### LINKED ISSUE #931: [Bug Report] Images in edits doesn't have correct aspect ratio

gimmeliina:
Updated image styles to use width and height instead of max dimensions.

This should finally fix #931 (related: #1030)
- Removed max-width and max-height, so Safari doesn't get confused about aspect-ratio.
- After that, correct dimensions is calculated 
- .Image-image fills 100% of it's container

### DIFF
--- a/frontend/src/components/image/styles.scss
+++ b/frontend/src/components/image/styles.scss
@@ -1,12 +1,10 @@
 .Image {
   position: relative;
   display: flex;
-  max-height: 100%;
-  max-width: 100%;
 
   &-image {
-    max-width: 100%;
-    max-height: 100%;
+    width: 100%;
+    height: 100%;
     z-index: 1;
   }
 


frontend/src/components/image/styles.scss
```diff
@@ -1,12 +1,10 @@
 .Image {
   position: relative;
   display: flex;
-  max-height: 100%;
-  max-width: 100%;
```
InfiniteStash, line 5: Happy  to merge the PR, but can  you revert the max-height/max-width removal? From my testing it's not necessary for Safari, and it screws up various things.

gimmeliina:
Fixed in #1160

PR #1151, opened 2026-06-05
Remove `margin-right` from gender icons

peolic:
It seems that is no longer needed, and just adds additional space between the icon and text, that was not there before.

<details>
<summary>Performer Page</summary>

before-performer<br>after-performer
</details>
<details>
<summary>Scene Page</summary>

before-scene<br>after-scene
</details>

Possibly changed in [Font Awesome v7.0.0](https://fontawesome.com/changelog#v7-0-0):
> Icon canvases now render at a consistent fixed width. To use the default rendering from v6, add our new fa-width-auto class.

### DIFF
--- a/frontend/src/pages/scenes/styles.scss
+++ b/frontend/src/pages/scenes/styles.scss
@@ -1,9 +1,3 @@
-.fa-mars,
-.fa-venus,
-.fa-transgender {
-  margin-right: 0.25rem;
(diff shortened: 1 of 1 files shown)


peolic:
This may have been a mistake.
I should have tested the other gender icons, they look worse.
Edit cards:
image

PR #1160, opened 2026-06-09
Switch from image carousel to lightbox

### LINKED ISSUE #777: [Feature] Link to Full Resolution Image
Previously, resized images were listed incorrectly in the edit queue, using their resized resolution instead of the full resolution. This bug was described in issue #662 and has since been resolved.

However, full resolution images are still only easily accessible through peolic's userscript. In the vanilla presentation of Stash-Box, only the resized image is displayed throughout the UI, including in the edit queue when the full resolution is listed. This makes it more difficult for contributors to restore a full-size image when it has been overwritten by the smaller resized image, an unfortunately common occurence.

Incorporating many of the features of peolic's userscript would go a long way. I went into more detail with various screenshots in #662 describing how it works. The visual language used by peolic's userscript is a bit counter-intuitive and inconsistent for me though. Sometimes the resolution labels are hyperlinks, sometimes they aren't. When an image has been resized an `(R)` hyperlink will appear throughout the UI, but clicking it opens the full-sized image and not the "resized" image that the "R" references. Also the resolution label will only hyperlink to an image with that listed resolution when it hasn't been resized. Instead, clicking the hyperlink on a label of `1920 x 1080` will actually open the image resized down to 1280 x 720 instead with the full-sized image is hidden behind the `(R)` hyperlink. These shortcuts and inconsistencies aren't a huge deal when only a small number of power users have the userscript, but they will need to be addressed if that functionality is added to stash-box itself. In order for new and less active users to understand what's going on, we also may have to spell it out more explicitly with labels like `Resized from <res>` or `Original size is <res>`.

Essentially, the remaining functionality exclusive to the userscript includes the following:
1. **Turn the resolution labels in the edit queue into hyperlinks**
    - Clicking `1920 x 1080` should open the full image in a new tab. Right now, you have to right-click the image and it only opens the resized 1280 x 720 image instead.
2. **Add a link to the full-size image in the scene details pages**
    - For me, the `(R)` hyperlink from the userscript is counterintuitive since it doesn't link to the "resized" image, so I'd change that to something else. Maybe an `(F)` for "full" or an `(O)` for "original"? A separate hyperlink for the smaller resized image is not needed since it can be easily opened with a right-click on the displayed image.
3. **Adds resolution labels and hyperlinks to the performer image carousel**
    - Again, the userscript's labels feel counter-intuitive here for me. The resolution label here is not a hyperlink, which is fine if we're adding a separate hyperlink for the full-size copy of the image. However, that wouldn't be needed if the resolution label itself was a hyperlink. Also a separate hyperlink for the smaller resized image is still not necessary here since we can easily right-click the displayed image.
4. **Adds resolution labels and hyperlinks to the "Images" tab of the edit forms**
    - The same visual language established in the earlier locations should be mirrored here as well. Resolution labels should be listed for each image and if they're also hyperlinks to open in a new tab, then the linked image should be the full-size image that matches the resolution label.

It's possible that more comprehensive links and labels to make stash-box's image resizing more transparent would be too confusing to the average user. If that's the case, then maybe some or all of these additions could be hidden from users without the EDIT role, or we could leave it exclusive to power users by reserving that functionality for the userscript. However, the current situation has been confusing and sometimes frustrating for many contributors who are unaware of stash-box's image resizing feature.

InfiniteStash:
Replaces the useless carousel with a lightbox that includes a thumbnail grid, so it's easier to actually see the performer images at a glance.

Also adds the lightbox to scene images, and edit change rows so it's possible to see images in full size.

Drafted with claude.

Resolves #777


slick-daddy:
This choice really broke the way I use stashdb. I think arrows were a nice option to browse performer images quickly. Now it requires to click on main image and then navigate between other images. I would prefer if both options kept or presented as an option.

PR #1182, opened 2026-06-25
Add optional TwelveLabs Marengo embedding integration

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

This PR adds an **opt-in** TwelveLabs Marengo embedding integration to augment stash-box's existing perceptual-hash (pHash) video matching.

### What it adds
- A small `internal/embedding` package with a stdlib-only TwelveLabs client (no new third-party dependency) that computes 512-dimension Marengo embeddings via the `/embed` endpoint, plus a `Cosine` similarity helper.
- A `twelvelabs` config block (`api_key`, `model_name`, `base_url`) wired through viper using the same `,squash` nested-struct pattern as the existing `s3`/`autocert` config, and a `GetTwelveLabsConfig()` accessor that returns `nil` when no key is set.
- README documentation for the new config keys and a short section explaining the integration.

### Why it helps
pHash reliably matches near-identical encodes but misses content that has been re-encoded, cropped, or otherwise visually altered. Marengo produces multimodal embeddings in a single space shared by text and video, so visually-similar scenes can be ranked by cosine similarity and text queries can be matched against video embeddings — a complementary signal for duplicate detection and search.

### Opt-in / non-breaking
The integration is inactive unless `twelvelabs.api_key` is configured. No defaults change and existing pHash behaviour is untouched. This PR adds the client + config seam; it deliberately does not alter any existing matching path.

### Testing
- `go test ./internal/embedding/` — table-driven unit tests for response parsing and cosine similarity (no network).
- A live test (`TestTextEmbeddingLive`) gated on `TWELVELABS_API_KEY`, skipped in CI without credentials. Ran it against the real API and confirmed a 512-dim vector is returned.
- `gofmt` and `go vet` clean on changed files; `go build ./internal/...` passes.

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,31 @@ There are two ways to authenticate a user in Stash-box: a session or an API key.
 | `autocert.domain` | (none) | The domain to generate certificates for.|
 | `autocert.email` | (none) | A valid email. Will be submitted to Let's Encrypt, but otherwise not made public. |
 | `mod_audit_retention_days` | 30 | Number of days to retain audit logs of moderator actions. Set `0` to disable. |
+| `twelvelabs.api_key` | (none) | API key for the optional [TwelveLabs](#twelvelabs-embeddings-optional) Marengo embedding integration. Leave empty to disable. |
+| `twelvelabs.model_name` | `marengo3.0` | TwelveLabs embedding model to use. |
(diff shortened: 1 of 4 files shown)


InfiniteStash:
You clearly do not understand what stash-box is.

PR #1209, opened 2026-08-20
Optimize slice comparison utility

slick-daddy:
The existing `SliceCompare` performed repeated linear membership scans which is O(n*m) and is hot in edit paths that compare aliases/images/tags, so lowering complexity will reduce CPU on those paths.

### Description

- Replace repeated `Includes` linear scans in `SliceCompare` with prebuilt set maps via a new `makeSet` helper to achieve O(n) lookups and reduce overall work in comparisons in `pkg/utils/slice_compare.go`.
- Add small maps (`addedSet` and `missingSet`) to suppress duplicates while preserving the original output order semantics of `SliceCompare`.
- Add a benchmark `BenchmarkSliceCompare` in `pkg/utils/slice_compare_test.go` exercising two 1,000-item slices with partial overlap to guard future regressions and measure improvements.
- Keep `Includes` available for callers that rely on it while centralizing set construction in `makeSet` for reuse and readability.

Drafted with Codex.

Edits usually don't include many objects, so practical change can be limited.

### DIFF
--- a/pkg/utils/slice_compare.go
+++ b/pkg/utils/slice_compare.go
@@ -11,17 +11,43 @@ func Includes[T comparable](arr []T, against T) bool {
 }
 
 func SliceCompare[T comparable](subject []T, against []T) (added []T, missing []T) {
+	againstSet := makeSet(against)
+	subjectSet := makeSet(subject)
+	addedSet := make(map[T]struct{}, len(subject))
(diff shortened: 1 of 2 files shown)


InfiniteStash:
Thanks! I removed the benchmark and tweaked a bit.

Issue #1205, opened 2026-08-20
[Bug Report] Low quality Performer image prioritization


sleetx:
The intent of #1089 was to prioritize a 2:3 image aspect ratio. The issue with  this change is that in some cases, very low resolution photos are chosen over more suitable ones. Aspect ratio cannot be the sole method of identifying a quality image.

For example, a 200x300 image (a perfect 2:3 ratio) gets prioritized over a 1200x2000 image. This is too extreme. There needs to be a reasonable cut off where the aspect ratio calculation is secondary to the resolution of a portrait image if there is a major difference between them.

tj2222:
Related issues: #161, #203, #237, #436

Issue #436, opened 2022-06-15
[Feature] Image voting system


echo6ix:
**Is your feature request related to a problem? Please describe.**
Image size and other file properties don't necessarily dictate quality or relevance of an image. Relevance can change over time, such as a more recent image.

**Describe the solution you'd like**
A simple voting system for performer images (and hopefully studio images if #279 is implemented), where users can vote on any image with an up-vote or down-vote. The image with the highest up-votes becomes the primary image.

**Additional context**
See screenshot from TheMovieDB
![image](https://user-images.githubusercontent.com/37937507/173960073-18d27e8c-431c-4066-b143-02a834015d0e.png)

**Note**
I realize this has been requested on #161 but I felt the issue required more fleshing out and a more relevant title

Issue #237, opened 2022-01-03
[RFC] Performer Image Categorization


AdultSun:
## Scope
There is a significant desire to better organize performer images on stash-box, particularly by allowing images to be categorized by crop, pose, and state of dress.

## Problem
Stash-Box performer images are currently organized by resolution only, basically serving the highest resolution image first followed by the rest in descending order. More popular performers can have many images and our StashDB guidelines encourage a diverse set for each performer representing various crops, poses, states of dress, and general "looks" from the performer's career. Typically, users will prefer a particular type of performer image for their personal use and are often left navigating stash-box's image carousel one-by-one until they find the "best" image. Currently, the only way to set a default image for a particular performer in stash-box is by making sure the preferred image has a higher resolution then the rest of the set. This is an imperfect work-around that requires maintaining a hard cap on image resolution for each performer rather than a true solution.

##  Proposal
The implementation of image categories could help to alleviate these issues. If individual images can be tagged with categories, users could set their own preferred defaults within both stash-box and their personal stash. A convenient system for categorization is already in use in the [actress-images repo](https://gitgud.io/Trizkat/actress-pics/-/tree/master) hosted by Trizkat. Though the collection is exclusively female performers (outside of a small number of AFAB non-binary performers), their system appears to be neutral enough to be suitable for performers of all genders. [In their own words](https://gitgud.io/Trizkat/actress-pics/-/blob/master/SCHEMA.md), with some light editing: 

### **Crop**
How much of the performer is visible?
- **Face**: Above the bust, and up
- **Bust**: Somewhere between the navel and bust, and up
- **Torso**: Above hips, below navel
- **3Q**: Three-quarters view, about mid-thigh, and up
- **Sit**: When you see the person sitting
- **Kneel**: Where you can see the knees touch the ground
- **Wide**: For landscape images where the subject is in some horizontal position
### **Pose**
How is the performer positioned in relation to the camera?
- **Portrait**
  - **Front**: Mostly facing towards the camera
  - **Back**: Mostly facing away from the camera
  - **Side**: Mostly facing perpendicular to the camera
- **Landscape**
  - **Back**: On their back
  - **Fours**, On all fours
  - **Front**: Mainly a frontal view and not covered by previous
### **State of Dress**
What is the performer wearing, if anything?
- **Non-Nude**: Covered up, or can't be determined
- **Topless**: Defined by if areola and nipple is uncovered
- **Nude**: Exposed bust and pubic region
### **Set**
[...]

Issue #212, opened 2021-12-03
[Feature] Add cropperjs to the add image section


echo6ix:
It might further improve the quality of uploaded pics by including [cropperjs](https://github.com/fengyuanchen/cropperjs) directly into the upload image section.

The cropperjs can be made to:
- Be locked into a 3:2 aspect ratio
- The cropper has a grid, and I'm sure it can be improved to suit the type of images Stashbox wants, such as a head guide that the uploader can use to fit their image into the guide like this: ![Screenshot 2021-12-03 002451](https://user-images.githubusercontent.com/37937507/144550139-af43e346-3fae-4701-b402-d7704c793def.png)
- Depending on the type of headshot the user selects, 1/4 headshot, 1/2 headshot, etc referenced here #211, the above guide will change to suit the headshot style.

PR #1214, opened 2026-08-24
Performer image labeling

Maista6969:
This PR provides two new features: performer images can be labeled and dated, and users of a stash-box can express their preferences for which images they'd like to see both when browsing and when scraping performers.

As discussed in #237 I opted for a closed taxonomy of types for the primary reason of compatibility: having a canonical set of labels for images means that users can use their preferences across multiple stash-boxes. Having users configure their preferences on the stash-box itself serves to simplify the rollout: if we had to rely on users configuring their preference in Stash then we'd need to coordinate releases, have Stash send an ordering whenever it scrapes a performer and also be more careful about backwards compatibility.

The taxonomy is intentionally wide because stash-box administrators can choose to disable the types they do not want to use for their labeling (maybe you think 3-quarters plus is excessive), and even disable entire groups (like dress because your instance doesn't have a lot of nude performers). When a label is disabled nothing is deleted: it is still present on any images that had it before the label was disabled, it is simply "inert" for the purposes of sorting images. Re-enabling the type brings the label back for all purposes.

There is an exception to the user-preferred ordering of images and that's search results, mostly because of their size: at such a small scale I believe we should prefer headshots since they stay readable at much smaller sizes. Performer images in search results will always use a face crop if the performer has one, otherwise it falls back to whatever it would have gotten before this feature was added. This affects the performer field in scene edit forms and the site-wide search in the top right corner of the app.

I believe this resolves:
- #237 - the primary motivation for this PR
- #161 - instead of making people vote for images this PR provides the users with their own personalized view, hopefully preventing (or at least mitigating) edit wars over which image is the "best" for a given performer
- #203 - this PR does not add aspect ratio as a label for images because that can be derived from the image dimensions themselves
- #211 - this PR lays the groundwork for this but would be further enhanced by the addition of something like the ["image bible"](https://www.themoviedb.org/bible/image) that TMDB has
- #1205 - performer images are now sorted by their labels, then their date, then their aspect ratio and finally their height as a tiebreaker

## For the reviewer(s):

This is a lot of code to review, so I've split the work into commits that can (hopefully) be reviewed in order. The very first commit adds new entries to `.gitattributes` to help GitHub know which files are autogenerated so that they'll be collapsed in the diff view by default. This won't help _this_ PR but for future work it should be helpful.

I've added tests, integration tests and end-to-end playwright tests for this feature, and while they're bot-assisted I've seen each of them fail in my development environment so at least there are no duds in there.

## AI Disclosure

I had Claude Opus look over the plan and assist with writing tests for this PR

### Follow-up work

- #212 - we can build a cropping UI on top of the "crop type" labels in this PR: if a user chooses the face crop label, we can immediately overlay a template with guidelines that help them crop the image to fit a common standard
- #436 - if we'd still like to add image voting that can easily be inserted into the sorting process


Maista6969:
Going to try reopening this as a stacked pull request to learn how they work on GitHub 🙂️

PR #1218, opened 2026-08-26
Fix accessibility and UX defects

slick-daddy:
#### 1. Auth redirect called during render (`Login.tsx`)

`if (isAuthenticated) navigate("/")` ran in the component body,
triggering React updates during render and risking a loop while auth
state resolves. Replaced with a declarative `<Navigate to="/" replace />`
returned after all hooks.

#### 2. react-select contrast failures (`App.scss`, `studioSelect/styles.scss`)

- Selected values and typed input text were forced to `color: black` on
  the dark `$secondary` control (~2.3:1). Now pinned explicitly to
  `$text-color` (~9.9:1).
- Multi-value chips paired near-white text with a near-white chip
  background; now `$dark-text` on the existing `$muted-gray` chip (~9.6:1).
- `StudioSelect`'s parent-studio label used `rgba(black, 0.5)` inside the
  value container, rendering near-invisible; dropped the override so it
  inherits the legible `$text-muted`.

All select surfaces (filters, forms, search) now meet WCAG AA text contrast.

#### 3. Deleted images invisible in edit diffs

Removed images were rendered as srcless `Deleted` tags, which
browsers draw as zero-width boxes — edit diffs showed blank gaps where
removed images should appear. Added a shared `DeletedImage` fragment
(muted tile, X icon, "Deleted" label, matching the existing
empty-thumbnail language) and used it in `ImageChangeRow` and
`AmendableImageChangeRow`.

#### 4. Global 1210px viewport lockout (`theme.scss`, `App.scss`)

`body { min-width: 1210px }` — carried over from the Bootstrap 5
migration — forced horizontal scrolling of the entire app on any narrower
viewport. Removed it: the navbar already wraps (`react-bootstrap`
defaults to `.navbar-expand`), grids auto-fill, and filter rows
flex-wrap. The login prompt's fixed `960px` width became `max-width` so
auth pages no longer overflow as the first screen seen at smaller sizes.
Wide tables can still overflow their own page; a full responsive pass is
out of scope here.

### DIFF
--- a/frontend/src/App.scss
+++ b/frontend/src/App.scss
@@ -37,6 +37,11 @@
   color: $text-color;
   cursor: pointer;
 
+  .react-select__single-value,
+  .react-select__input {
+    color: $text-color;
+  }
+
   .react-select__multi-value {
     background-color: $muted-gray;
     color: $dark-text;
--- a/frontend/src/App.scss
+++ b/frontend/src/App.scss
@@ -60,7 +60,7 @@ div.react-select__menu {
 
 .LoginPrompt {
   height: 70vh;
-  width: 960px;
+  max-width: 960px;
   margin-left: auto;
   margin-right: auto;
   display: flex;
--- a/frontend/src/App.scss
+++ b/frontend/src/App.scss
@@ -37,14 +37,9 @@
   color: $text-color;
   cursor: pointer;
 
-  .react-select__single-value,
-  .react-select__input {
-    color: black;
-  }
-
   .react-select__multi-value {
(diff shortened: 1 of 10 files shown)


InfiniteStash:
Please split unrelated fixes up into separate PRs. Makes them much easier to review and merge.

slick-daddy:
I tried to keep the commits in atomic form. Comments are well written too but if you still find it hard to review, of course I can split. How do you want me to split?

InfiniteStash:
Based on the PR description I would expect four PRs.

PR #1230, opened 2026-09-02
Wrap debounce in useMemo to prevent recreation on every render

slick-daddy:
## Description
debounce(setParams, 200) was called directly in the render body of 6 components, creating a brand new debouncer on every render. This resets the debounce timer and defeats the debounce mechanism entirely.

Wrap each in useMemo with [setParams] dependency. In Performers.tsx and Users.tsx, move the hook before the early return to satisfy Rules of Hooks.

## Cleanup on unmount

Added a `useEffect` cleanup to cancel the pending debounced call when the
component unmounts, preventing a potential state update on an unmounted
component:

```tsx
useEffect(() => () => debouncedHandler.cancel(), [debouncedHandler]);
```

Applied to all six components touched by the `useMemo` fix above:
`TagList`, `Performers`, `ScenePairings`, `Studios`, `StudioPerformers`,
and `Users`.

### DIFF
--- a/frontend/src/components/list/TagList.tsx
+++ b/frontend/src/components/list/TagList.tsx
@@ -1,5 +1,5 @@
 import { debounce } from "lodash-es";
-import type { FC } from "react";
+import { type FC, useMemo } from "react";
 import { Card, Form, Row } from "react-bootstrap";
 import { Link } from "react-router-dom";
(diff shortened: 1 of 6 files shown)


InfiniteStash:
Thanks, but closing in favor of https://github.com/stashapp/stash-box/pull/1246

PR #1243, opened 2026-09-03
Add Circumcised and Penis Length fields for performers

he69man69:
# Add Circumcised and Penis Length fields for performers

## Summary

Stash-box models female-specific body attributes (breast type, cup/band/waist/hip size) on the Performer entity, but has no equivalent for male/transfemale/non-binary/intersex performers. Stash already supports circumcision status and penis length, and this closes that gap so stash-box can store and serve the same data, keeping edit history, voting, and merge-conflict resolution consistent with every other performer attribute.

- Adds two new nullable Performer fields:
  - **Circumcised** — enum, `CUT` / `UNCUT`.
  - **Penis length** — `Int`, centimeters (matches the existing convention of height/waist/hip being plain `Int`, not float)
- Threaded through the full stack: GraphQL schema, Postgres migration, sqlc queries, edit-diff/voting system, merge-conflict UI, performer drafts (for scraper/Stash submissions), and the performer edit form and detail page.
- **Visibility is gender-aware**, mirroring how breast attributes are already gated, but as an independent rule so transfemale performers can carry both breast and penis attributes, and transmale performers get neither:

  | Gender | Breast attributes | Penis attributes |
  |---|---|---|
  | Female | ✅ | ❌ |
  | Male | ❌ | ✅ |
  | Transgender Female | ✅ | ✅ |
  | Transgender Male | ❌ | ❌ |
  | Non-binary / Intersex / unset | ✅ | ✅ |

- No search/filter support was added (stored + displayed only, matching the smaller scope of most measurement fields).
- No backend gender enforcement, matching existing precedent (breast_type has none either).
- Bumps `schemaVersion` in `internal/database/database.go` from 75 to 76 to accomodate the new DB fields.

## Screenshots

Screenshot 2026-09-03 at 10 28 35

Screenshot 2026-09-03 at 10 30 01

## Known issues in testing

- Stash app doesn't submit these penis attributes via Draft submissions, so this will have to be handled separately.

## Test plan

- [x] `go build`, `go vet`, `go test` across affected backend packages
- [x] `tsc --noEmit`, `biome check`, full `vitest` suite (373 tests) on the frontend
- [x] Manually verified end-to-end against a local stash-box instance connected to a local Stash instance:
  - Created/edited performers of each gender and confirmed field visibility matches the table above
  - Confirmed pending-edit diff view, merge-conflict UI, and performer detail page render the new fields correctly
  - Confirmed performer draft submission from Stash round-trips the new fields once Stash sends them
  - Confirmed values clear correctly when switching a performer's gender to one where a field group no longer applies


he69man69:
Closing, two more issues to fix.

PR #1256, opened 2026-09-06
Dedup cron cleanDrafts with semaphore

slick-daddy:
## What I did
- #10: `internal/cron/cron.go` — added `sem.TryAcquire` + `defer Release` to `cleanDrafts()`

## Why needed
- Only `processEdits` had dedup guard; other cron tasks could overlap/run twice.

## Impact
- 5 lines; prevents duplicate `cleanDrafts` runs; zero behavior change for single-run case.


InfiniteStash:
This is meaningless, there is no chance of draft cleaning running on top of each other, and it wouldn't matter if they did.

PR #1258, opened 2026-09-06
Batch query param pre-alloc

slick-daddy:
## What I did
- #12: `internal/queries/helpers.go` — added `QueryIDs` with `make(..., len(ids))` pre-alloc.

## Why needed
- Query parameter slices built via append; capacity known from input.

## Impact
- 6 lines; fewer allocs per batch query.


InfiniteStash:
This literally does nothing.

PR #1259, opened 2026-09-06
Batch bm25Search plan setup once

slick-daddy:
## What I did
- #4: `internal/service/performer/service.go` — `bm25Search` plan setup wrapped in `sync.Once`.

## Why needed
- Each search call ran `SET LOCAL plan_cache_mode`; once is sufficient per process.

## Impact
- 7 lines; removes repeated DDL overhead.


InfiniteStash:
This is incorrect.

PR #1261, opened 2026-09-06
Lazy PerformerInfo/ScenePairings

slick-daddy:
## What I did
- Frontend #8: `frontend/src/pages/performers/Performer.tsx` — `PerformerInfo` + `ScenePairings` loaded via `React.lazy`
- Added `React.Suspense` to make them actually work
## Why needed
- Heavy sub-components bundled eagerly; split by route reduces JS load.

## Impact
- Lazy load only when tab rendered


InfiniteStash:
This is incorrect. PerformerInfo is always loaded, and ScenePairings is only mounted on demand.

PR #1268, opened 2026-09-09
Update github actions

InfiniteStash:


.github/workflows/build.yml
```diff
@@ -30,36 +30,29 @@ jobs:
...
-
-    - name: Install Node
-      uses: actions/setup-node@v4
-      with:
-        node-version: '24'
 
     - name: Install PNPM
       uses: pnpm/action-setup@v6
       with:
         version: 11
```
slick-daddy: pnpm 12 is stable and it is a rewrite in Rust, it makes sense to upgrade: https://pnpm.io/blog/releases/12.0

.github/workflows/frontend-lint.yml
```diff
@@ -21,25 +21,19 @@ jobs:
...
-      - name: Cache node packages
-        uses: actions/cache@v4
-        env:
-          cache-name: cache-node_modules
+      - name: Install Node
+        uses: actions/setup-node@v7
         with:
-          path: frontend/node_modules
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('frontend/pnpm-lock.yaml') }}
+          node-version: '24'
```
slick-daddy: Node 26 is stable as well but gains are not much. So it is up to us: https://nodejs.org/en/blog/release/v26.8.2

Issue #1141, opened 2026-06-02
[Feature] Additional metadata for male performers


InfiniteUsernames:
**Is your feature request related to a problem? Please describe.**
Male performers are missing basic metadata fields in StashDB, such as penis size and circumcision status. Stashapp does have these fields, but they are missing in the database.

While penis size and circumcision status are the most basic, it would additionally be useful to have body type markers (similar to what is available for women in breast size and waist size). For male performers, we assume it would make sense to use nomikers like "twink" or "bear". That said, I assume this is less objective and might be harder to implement than something like penis size and circumcision status

**Describe the solution you'd like**
Add at least the penis size and circumcision status fields to StashDB

For the body types, perhaps this can be associated with tags to give some degree of flexibility?

**Additional context**
See also this discussion: https://discourse.stashapp.cc/t/metadata-for-male-performers/6561

discourse-stashapp:
This issue has been mentioned on **Stash Forum**. There might be relevant details there:

https://discourse.stashapp.cc/t/metadata-for-male-performers/6561/16

cgnz2002:
+1 on this

he69man69:
Copying over to relevant issue:

Proposal on handling this:

- Genitals (new field, not always shown), which is used in validation for other attributes (Vulva/Penis attributes)
- Genders (existing model, not touching anything there, but used to gate genital attributes):
  - Male: Has Penis inferred[^1] (no additional dropdown), and gets Penis attributes. Has N/A Breast attributes and no frontend options (as it does today). **Net effect: Gets two new boxes of Circumcised and Penis length, matching Stash app.** 
  - Female: Has Vulva inferred (no additional dropdown), and gets Vulva attributes if that comes down the line so the data model can be ready. Has Breast attributes (as it does today). If more attributes are recommended in how the Vulva looks, those can get shown (thinking of Innie Pussy and Outie Pussy tags, Natural pre-selected vs Artifical for trans, etc). **Net effect: No new boxes.**
  - Trans/Intersex/NB performers: Gets Genital/Breast Type dropdowns (with defaults selected) to describe either case as needed.
    - Transmales: Neither are selected by default, since majority of porn showcases a person with breasts removed and a vulva.
    - Transfemales: Both are selected by default, since majority of porn showcases a person with breasts and a penis.

Net DB effect on new fields (with no effect on existing field options):
[...]

Issue #784, opened 2024-05-17
Add cut/uncut and penis length to stastdh


sarimsalman93:
While it is already in the stashapp, the circumcision status and penis length fields are not available yet in stashdb. Would be nice to add it.

echo6ix:
Partial duplicate of and related to #206

Issue #206, opened 2021-11-30
[Feature] Add penis size, for male and trans performers


ghost:
Should function similarly to measurement fields in that it's hidden for "incompatible" genders.

baloodusudouest:
Agree!

PR #1269, opened 2026-09-10
Add genital attributes to performers

he69man69:
# Add genital attributes to performers

Adds a "bottom half" counterpart to the existing breast attributes, so performers can be described as they appear on screen. Off by default, and instances opt in explicitly via config.

Ties into feature 206 on https://github.com/stashapp/stash-box/milestone/1

## New fields

_Edit: Augmented Vagina is now Constructed Vagina_

| Field | Type | Storage |
|---|---|---|
| `genitals` | `GenitalEnum`: `CIR_PENIS`, `UNCIR_PENIS`, `NAT_VAGINA`, `CONS_VAGINA` | nullable `varchar(20)` |
| `penis_length` | `Int` (centimetres) | nullable `integer` |

"Unknown" is the null / unset state (the default for every existing and new performer).

Both fields are wired through the same paths as `breast_type` / the measurement fields: `Performer`, `PerformerCreateInput` / `UpdateInput` / `EditDetailsInput`, `PerformerEdit`, the edit / voting / merge, and drafts (`genitals` only, as a free-text string, mirroring `breast_type`; `penis_length` is not a draft field, mirroring `hip_size`). 

## Admin opt-in: `enable_genital_attributes`

New config key (default **false**), exposed on `StashBoxConfig` via `getConfig`. Same pattern as `require_tag_role` (config file only; stash-box has no settings UI).

When disabled, the frontend hides the feature everywhere via a `useConfig()` gate: the performer form field, the detail-page rows, and the edit-diff / amendable-edit change rows. `PerformerForm` also omits `genitals` / `penis_length` from the submitted edit when disabled, so an instance that turns the feature off never wipes values that were set while it was on. Instances that don't want this or whose content never depicts genitals (e.g. JAV) just never turn it on via config.

## Migration

`76_performer_genitals.up.sql` adds the two columns, and `schemaVersion` in `internal/database/database.go` is bumped from 75 to 76.

## Form behaviour

The Genitals `<select>` is guided by gender (every list also has a leading "Unknown"):

| Gender | Options |
|---|---|
| Male | Circumcised penis, Uncircumcised penis |
| Female | Natural vagina, Constructed vagina |
| Transgender male/female, intersex, non-binary, unset | all four |

- **Penis length** is shown when the selected genitals value is a `*_PENIS`. Before anything is selected it falls back to gender (shown for Male only, where circ/uncir can be unknown but penis_length can be entered).
- Gender only changes which options are offered. A stored value is not auto-cleared or force-set, and there is no `useEffect` mutating the field. Editing an unrelated attribute therefore does not produces a `genitals` / `penis_length` diff. (Deliberately unlike `breast_type`, which force-sets `NA` for male performers. `genitals` has no sentinel value and the column is nullable. Existing `breast_type` behaviour is untouched.)

## Display

`genitals` and `penis_length` show on the performer detail page (all genders) and as change rows in the edit card and amendable-edit card.

## Notes for reviewers

- The `genitals` enum is kept deliberately simple to address the three feature requests (#206, #784, #1141), which request more info on male/trans performers. This enum can be expanded if there's more demand for different types.
- `penis_length` has no backend range validation (matches `hip_size`); the form enforces 1–45 cm. This is to prevent a height getting entered there, and a performer getting a 6' penis :)
- No `.down.sql` for the migration (consistent with recent migrations in the repo).

## Tests

- **Backend**: `genitals` / `penis_length` added to the shared performer create/edit fixtures and `verify*` helpers (same approach as `breast_type` / measurements); `TestApplyModifyPerformerEnumFields` exercises a modify round-trip; `extension_edit_details_test.go` covers the diff logic.
- **Frontend**: new cases in `PerformerForm` (per-gender option lists, penis-length visibility, feature-disabled hiding), `diff`, `merge`, and `renderPerformerDetails` tests. A `configGenitalsEnabledMock` drives the enabled state.


he69man69:
Adding some some before/after stats blocks to show the visual effect of the update:

### Female
#### Before
female-before

#### After
female-after

### Male
#### Before
male-before

#### After
male-after

### Trans-Female
#### Before
tf-before

#### After
tf-after

### Trans-Male
#### Before
tm-before

#### After
tm-after

### Non-Binary
#### Before
nb-before

#### After
nb-after

PR #1272, opened 2026-09-14
Add an id criterion to SceneQueryInput

alextomas955:
Adds `id: MultiIDCriterionInput` to `SceneQueryInput`, so a client can include
or exclude specific scenes by id. There is currently no scene-id field on that
input in either direction.

The case this serves is "this studio's catalogue minus the scenes I already
have". Today a client has to fetch the catalogue and subtract locally, which
means `per_page` and `count` describe the unfiltered set and paging has to be
redone client-side.

`scenes.id` is a direct column and is never null, so this follows the `studios`
filter in the same file: `EQUALS`/`NOT_EQUALS` on a single value,
`INCLUDES`/`EXCLUDES` on the list, and an error on anything else. `IS_NULL` and
`NOT_NULL` don't apply to a primary key, so they're rejected rather than
silently ignored.

The `TRENDING` sort pushes a `LIMIT` into its subquery only when no other
filter is set, so the new field is added to that check alongside the existing
ones. Without that, a trending query combined with an id filter would limit
before filtering and return the wrong page.

### On the size of the id list

I have not added a cap, because where one belongs is your call. There is a hard
ceiling already: each list element binds as its own SQL parameter, and
Postgres's extended protocol stops at 65535. Measured against a local instance:

| ids | result |
|---|---|
| 60,000 | ok, 160 ms, 2.3 MiB request body |
| 65,534 | ok |
| 65,535 | `extended protocol limited to 65535 parameters` |

The query binds one parameter for `scenes.deleted`, which is why the last id
that fits is 65,534; combining the filter with others lowers it further. Past
that point the failure is a clear error rather than a wrong answer. Happy to
add a documented limit if you want one, and to put it wherever you'd prefer.

**Validation**: ran `make generate` (gqlgen, graphql-codegen, sqlc 1.29.0) and
committed the output; sqlc produced no diff. Built the binary and ran it
against Postgres in Docker, then exercised every modifier over HTTP. For the
trending path I checked the emitted SQL with `log_statement=all`: without an id
filter the subquery carries `LIMIT 10 OFFSET 0` inside, with one it does not
and the limit is applied outside, which is the intended behaviour.
`TestQueryScenesByID` covers `INCLUDES`, `EXCLUDES`, `EQUALS`, `NOT_EQUALS` and
four rejected modifiers. `make lint test it` passes.


InfiniteStash:
What's the use-case for this?

Generally speaking I would say it's not a good pattern to have users send the entire contents of their stash as a parameter for anyone to spy on.

